PR #2, opened 2016-12-05
renaming library.py to import_content.py

coreydaley:


coreydaley:
@bparees please review and merge

PR #6, opened 2017-01-17
updating import_content.py to work with yaml files

### LINKED ISSUE #7: Rename index.md to readme.md
Currently, the rendered content of the index.md pages does not show up in the community/official directory view. We probably need to rename this to readme.md so they show up in the directory view.

https://github.com/openshift/library/blob/master/community/
https://github.com/openshift/library/blob/master/official/

coreydaley:
Fixes #7

### DIFF
--- a/import_content.py
+++ b/import_content.py
@@ -29,28 +29,56 @@ def message(action, sub_action, text):
     """
     print('{:<10} | {:<17} | {}'.format(action, sub_action, text))
 
-def is_valid(data, data_type):
+def is_json(data):
+    """ Checks if the data is json
+
+    Args:
+        data: The data to check
+
+    Returns:
+        bool: True if the data is valid json, False otherwise
+    """
+    try:
+        json.loads(json.dumps(data))
+    except:
+        return False
+    return True
+
+def is_yaml(data):
+    """ Checks if the data is valid yaml
+
+    Args:
+        data: The data to check
+
+    Returns:
+        bool: True if the data is valid yaml, False otherwise
+    """
+    if data[0] == "{":
+        return False
+    try:
+        yaml.load(data)
+    except:
+        return False
+    return True
+
+def is_valid(data):
     """Checks the validity of a JSON or YAML document
 
     Args:
         data (string): the data to check
-        data_type (string): json|yaml
 
     Returns:
         bool: True if the data is valid, False otherwise
-        data: None if validity is False, otherwise a JSON or YAML object
+        data: None if validity is False, otherwise a Python dict object
 
     """
-    try:
-        if data_type == 'json':
-            loaded_data = json.loads(json.dumps(data))
-        elif data_type == 'yaml':
-            loaded_data = yaml.load(data)
-    except:
-        message('Error', 'validation', 'data is not valid ' + data_type)
+    if is_yaml(data):
+        return True, yaml.load(data)
+    elif is_json(data):
+        data = data.replace("\t", " ")
+        return True, json.loads(data)
+    else:
         return False, None
-    message('Success', 'validation', 'data is valid ' + data_type)
-    return True, loaded_data
 
 def append_to_index(source, folder, sub_folder, data):
     """ Appends information about a template or image-stream to the index file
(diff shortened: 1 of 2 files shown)


coreydaley:
@bparees ptal

import_content.py
```diff
@@ -29,28 +29,53 @@ def message(action, sub_action, text):
...
     """
-    try:
-        if data_type == 'json':
-            loaded_data = json.loads(json.dumps(data))
-        elif data_type == 'yaml':
-            loaded_data = yaml.load(data)
-    except:
-        message('Error', 'validation', 'data is not valid ' + data_type)
+    if is_yaml(data):
+        return True, yaml.load(data)
```
bparees: isn't all json valid yaml?

import_content.py
```diff
@@ -107,26 +132,37 @@ def fetch_or_retrieve_url(path):
...
+def write_cache_data(data, path):
+    """ Writes cached yaml data to a file at path
+
+    Args:
+        data (dict): dict data to write to the file
+        path (string): the path to write the file to, includes folder and file name
+    """
+    message('Writing', 'data to file', path)
+    with open(path, 'wb') as target_file:
+        target_file.write(bytes(data))
```
bparees: what's the motivation for having two methods to write to disk, one for yaml and one for json?  can we not use yaml everywhere?

import_content.py
coreydaley: Yes, except that our json files have tabs in them (doh!) and when you call yaml.load() on json with tabs in it, it throws an exception.  So it's safer to do it this way

luciddreamz:
In Ruby, the following lines of JSON were not converting properly to YAML:

    {
      "sampleRepo": "https://github.com/openshift/rails-ex.git",
      "description": "The exposed hostname that will route to the Rails service, if left blank a value will be defaulted.",
      "from": "user[A-Z0-9]{3}",
      "name": "ruby:2.3",
      "emptyDir": {},
      "exec": {
        "command": [
          "psql -h 127.0.0.1 -U ${POSTGRESQL_USER} -q -d ${POSTGRESQL_DATABASE} -c 'SELECT 1'"
        ]
      }
    }

How they convert in Ruby (incorrect):

    value: https://github.com/openshift/rails-ex.git
    description: The exposed hostname that will route to the Rails service, if left blank a value will be defaulted.
    from: user[A-Z0-9]{3}
    name: ruby:2.3
    emptyDir: {}
    exec:
      command:
      - psql -h 127.0.0.1 -U ${POSTGRESQL_USER} -q -d ${POSTGRESQL_DATABASE} -c 'SELECT 1'

Corrected conversion:

    value: "https://github.com/openshift/rails-ex.git"
    description: > The exposed hostname that will route to the Rails service, if left blank a value will be defaulted.
    from: "user[A-Z0-9]{3}"
    name: "ruby:2.3"
    emptyDir: "{}"
    exec:
      command:
      - "psql -h 127.0.0.1 -U ${POSTGRESQL_USER} -q -d ${POSTGRESQL_DATABASE} -c 'SELECT 1'"

coreydaley:
@bparees @luciddreamz ptal

luciddreamz:
@coreydaley so YAML/JSON source will be output as JSON?

coreydaley:
Correct

bparees:
(I haven't forgotten about this)

import_content.py
```diff
@@ -90,52 +118,41 @@ def replace_variables(string):
             message('Processing', 'variables', 'replacing ' + match + ' with ' + SVARS['vars'][re.sub(r'\{|\}', '', match)])
     return string
 
-def fetch_or_retrieve_url(path):
-    """ Fetches json data from a url and caches it locally for faster access
+def fetch_url(path):
+    """ Fetches json data from a url
```
bparees: json or yaml?

import_content.py
```diff
@@ -90,52 +118,41 @@ def replace_variables(string):
...
-                return "Error, invalid json detected", None
-            write_data_to_file(json_data, cache_path)
-            return None, json_data
-        else:
-            return req.status_code, None
+        return req.status_code, None
 
 def write_data_to_file(data, path):
-    """ Writes data to a file at path
+    """ Writes formatted json data to a file at path
```
bparees: so all content is ultimately converted to json even if it originated as yaml?  @luciddreamz were there any concerns with that?

coreydaley: That is correct, because of the double quote issue with writing the yaml files that @luciddreamz encountered, which is also present in python.  WHICH makes me think that if the string needs to be in double quotes that it should have escaped double quotes around it inside of the double quotes in yaml?  Would it be worth investigating that code that runs the command and see what it does to it?  i would assume that it is removing the quotes since there are quotes around each section of the command in the yaml...

bparees: my brain hurts.  but yes it sounds like we should test a yaml file that has some quoted strings like a command arg and see how it comes out.

luciddreamz: Just make sure my earlier example converts properly back from YAML to JSON:

    value: "https://github.com/openshift/rails-ex.git"
    description: "The exposed hostname that will route to the Rails service, if left blank a value will be defaulted."
    from: "user[A-Z0-9]{3}"
    name: "ruby:2.3"
    emptyDir: "{}"
    exec:
      command:
      - "psql -h 127.0.0.1 -U ${POSTGRESQL_USER} -q -d ${POSTGRESQL_DATABASE} -c 'SELECT 1'"

Should come out:

    {
      "sampleRepo": "https://github.com/openshift/rails-ex.git",
      "description": "The exposed hostname that will route to the Rails service, if left blank a value will be defaulted.",
      "from": "user[A-Z0-9]{3}",
      "name": "ruby:2.3",
      "emptyDir": {},
      "exec": {
        "command": [
          "psql -h 127.0.0.1 -U ${POSTGRESQL_USER} -q -d ${POSTGRESQL_DATABASE} -c 'SELECT 1'"
        ]
      }
    }

bparees: @coreydaley does it?

coreydaley: Yes.

bparees:
lgtm then.

PR #16, opened 2017-03-06
Adds bucharest-gold Node.js ImageStream

luciddreamz:
@coreydaley @bparees Please merge

Note: We may want to consider updating the import_python script to forcibly rename image stream files. At the moment, it names the file based on the name of the image stream. We may want to either add an option to force an explicit name, or simply name the file: technologyName-suffix.json.

### DIFF
--- a/community/README.md
+++ b/community/README.md
@@ -117,6 +117,10 @@ Path: community/mysql/templates/mysql-persistent.json
 Source URL: [https://raw.githubusercontent.com/openshift/origin/master/examples/image-streams/image-streams-centos7.json](https://raw.githubusercontent.com/openshift/origin/master/examples/image-streams/image-streams-centos7.json )  
 Docs: [https://github.com/sclorg/s2i-nodejs-container/blob/master/README.md](https://github.com/sclorg/s2i-nodejs-container/blob/master/README.md)  
 Path: community/nodejs/imagestreams/nodejs-centos7.json  
+### centos7-s2i-nodejs
+Source URL: [https://raw.githubusercontent.com/bucharest-gold/origin-s2i-nodejs/master/image-streams.json](https://raw.githubusercontent.com/bucharest-gold/origin-s2i-nodejs/master/image-streams.json )  
+Docs: [https://github.com/bucharest-gold/origin-s2i-nodejs/blob/master/README.md](https://github.com/bucharest-gold/origin-s2i-nodejs/blob/master/README.md)  
+Path: community/nodejs/imagestreams/centos7-s2i-nodejs-bucharest-gold.json  
 ## templates
 ### nodejs-example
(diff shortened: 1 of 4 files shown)


community/index.json
```diff
@@ -206,6 +206,12 @@
                 "name": "nodejs",
                 "path": "community/nodejs/imagestreams/nodejs-centos7.json",
                 "source_url": "https://raw.githubusercontent.com/openshift/origin/master/examples/image-streams/image-streams-centos7.json"
+            },
+            {
+                "docs": "https://github.com/sclorg/s2i-nodejs-container/blob/master/README.md",
+                "name": "centos7-s2i-nodejs",
+                "path": "community/nodejs/imagestreams/centos7-s2i-nodejs--bucharest-gold.json",
```
bparees: double dash looks wrong.  suffix probably doesn't need to include the dash prefix.

bparees: also i'm not a fan of reversing the naming scheme.

community/README.md
```diff
@@ -117,6 +117,10 @@ Path: community/mysql/templates/mysql-persistent.json
 Source URL: [https://raw.githubusercontent.com/openshift/origin/master/examples/image-streams/image-streams-centos7.json](https://raw.githubusercontent.com/openshift/origin/master/examples/image-streams/image-streams-centos7.json )  
 Docs: [https://github.com/sclorg/s2i-nodejs-container/blob/master/README.md](https://github.com/sclorg/s2i-nodejs-container/blob/master/README.md)  
 Path: community/nodejs/imagestreams/nodejs-centos7.json  
+### centos7-s2i-nodejs
+Source URL: [https://raw.githubusercontent.com/bucharest-gold/origin-s2i-nodejs/master/image-streams.json](https://raw.githubusercontent.com/bucharest-gold/origin-s2i-nodejs/master/image-streams.json )  
+Docs: [https://github.com/sclorg/s2i-nodejs-container/blob/master/README.md](https://github.com/sclorg/s2i-nodejs-container/blob/master/README.md)
```
bparees: is this docs link right?  i'd expect it to go to the bucharest repo.

community/nodejs/imagestreams/centos7-s2i-nodejs--bucharest-gold.json
```diff
@@ -0,0 +1,264 @@
+{
+    "apiVersion": "v1",
+    "kind": "ImageStream",
+    "metadata": {
+        "creationTimestamp": "2017-03-02T17:50:12.912Z",
+        "name": "centos7-s2i-nodejs"
```
bparees: i take it they decided to do this in their imagestream definition.  would love to know why.

luciddreamz:
Removed the extra dash in the suffix. Fixed the docs link.

coreydaley:
This might be better suited to being it's own top level element, similar to mailcatcher and swift.  That will put it into it's own directory, then you could remove the suffix.

bparees:
@coreydaley we decided we don't want 0923840 toplevel nodejs-foo directories so nesting them all under nodejs seems preferable.

PR #18, opened 2017-04-25
Add custom tags and allow the script to filter items based on tags

dinhxuanvu:
The new tags for online-starter and online-professional are added to selected
templates and imagestreams. The script is modified to accept the custom tags via
flag '-t' or '--tags' and custom tags are concatenated but separated by comma.

If custom tags are provided, all templates and imagestreams will be imported under
the folder with the name of the tag(s) and under /templates or /imagestreams
respectively. There are not further sub-directori(es).

The index.json and README.rd are also added into the folder(s) to provide further
information in case users need to locate the file(s) or source(s).

The script is now able to detect file collision and will append index numer to
file name to avoid file overwritten issue.

Updated the README to reflect new Online tags for YAML files and further details.

Signed-off-by: Vu Dinh <vdinh@redhat.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,15 +52,15 @@ Community templates and image streams are **not** provided or supported by Red H
 
 ### Running the Script
 
-    $ make import
-    
+    $ make import TAGS=<tag>,...
+
 ## Verifying Your Updates
 
     $ make verify
-    
+
 The `make verify` command runs the following checks:
  - verifies YAML syntax
- - verifies the Python script *(using pylint)* 
+ - verifies the Python script *(using pylint)*
  - verifies that make import has been run
 
 ## Contributing
@@ -89,11 +89,19 @@ That's it!  Your pull request will be reviewed by a member of the OpenShift Team
             regex: # (optional) matched against ['metadata']['name'] in the json file
             suffix: # (optional) suffix for the file that is created ex: ruby-<suffix>.json
             docs: # (optional) web address of the documentation for this image-stream
+            # Custom tags for OpenShift are added below if needed
+            openshift: # (optional) flag for OpenShift
+                - online-starter # (optional) flag for OpenShift Online Starter
+                - online-professional # (optional) flag for OpenShift Online Professional
         templates: # (optional) list of templates to process into the above folder
           - location: # (required) github url to a template or folder of templates in json format
             regex: # (optional) matched against ['metadata']['name'] in the json file
             suffix: # (optional) suffix for the file that is created ex: ruby-<suffix>.json
             docs: # (optional) web address of the documentation for this template
+            # Custom tags for OpenShift are added below if needed
+            openshift: # (optional) flag for OpenShift
+                - online-starter # (optional) flag for OpenShift Online Starter
+                - online-professional # (optional) flag for OpenShift Online Professional
 
 #### Variables
 
@@ -112,6 +120,8 @@ Listings in the **official.yaml** file will be created in a sub folder of the  *
 #### folder_name
 
 The **folder_name** is a sub folder which represents a logical grouping for a set of templates or image-streams in the top level **official** or **community** folders.
+
+For custom tags option, templates are imported into <tag>/templates/ directory while imagestreams are imported into <tag>/imagestreams/
 
 #### location
 
--- a/community.yaml
+++ b/community.yaml
@@ -3,6 +3,9 @@ data:
     templates:
       - location: https://raw.githubusercontent.com/luciddreamz/laravel-ex/master/openshift/templates/laravel-mysql.json
         docs: https://github.com/luciddreamz/laravel-ex/blob/master/readme.md
+        openshift:
+          - online-starter
+          - online-professional
       - location: https://raw.githubusercontent.com/luciddreamz/laravel-ex/master/openshift/templates/laravel-postgresql.json
         docs: https://github.com/luciddreamz/laravel-ex/blob/master/readme.md
       - location: https://raw.githubusercontent.com/luciddreamz/laravel-ex/master/openshift/templates/laravel-sqlite.json
@@ -27,6 +30,9 @@ data:
         docs: https://github.com/openshift-s2i/s2i-wildfly/blob/master/README.md
         regex: wildfly
         suffix: centos7
+        openshift:
+          - online-starter
+          - online-professional
   infinispan:
     imagestreams:
       - location: https://raw.githubusercontent.com/infinispan/infinispan-openshift-templates/master/imagestreams/infinispan-centos7.json
--- a/import_content.py
+++ b/import_content.py
@@ -14,10 +14,10 @@ import yaml
 SVARS = {
     "base_dir": os.path.dirname(os.path.dirname(os.path.realpath(__file__))),
     "sources": ['official', 'community'],
+    "tags": [],
     "vars": {},
     "index": {}
     }
-
 
 def message(action, sub_action, text):
     """ Outputs a formatted message to stdout
@@ -162,6 +162,25 @@ def write_data_to_file(data, path):
     target_file.write(json.dumps(data, sort_keys=True, indent=4, separators=(',', ': ')))
     target_file.close()
 
+def check_file_name(path, file_name, extension):
+    """ Check if the file already exists in the directory and return an valid file name
+
+    Args:
+        path (string): the path to write the file to, includes folder and file name
+        file_name (string): the file name
+        extension (string): the name of file extension (such as .json)
+
+    Returns:
+        file_name (string): The valid file name that can be written into
+
+    """
+    i = 1
+    name = file_name
+    while os.path.exists(path + name + extension):
+        name = file_name + "_" + str(i)
+        i = i + 1
+    return name
+
 
 def process_template(source, folder, location_list, template):
     """ Processes a template and writes it's data to a file in JSON format
@@ -178,15 +197,16 @@ def process_template(source, folder, location_list, template):
     if 'regex' in location_list:
         matches = re.search(r'(^' + location_list['regex'] + ')', template["metadata"]["name"])
     if matches or 'regex' not in location_list:
+        file_name = check_file_name(SVARS['base_dir'] + '/' + folder + '/templates/', template['metadata']['name'], '.json')
         index_data = {
             'name': template['metadata']['name'],
             'docs': location_list['docs'] if 'docs' in location_list else '',
             'source_url': location_list['location'],
             'description': template['metadata']['annotations']['description'] if 'description' in template['metadata']['annotations'] else '',
-            'path': source + '/' + folder + '/templates/' + template['metadata']['name'] + '.json'
+            'path': (source + '/' if source in SVARS["sources"] else '') + folder + '/templates/' + file_name + '.json'
             }
         append_to_index(source, folder, 'templates', index_data)
-        write_data_to_file(template, SVARS['base_dir'] + '/' + folder + '/templates/' + template['metadata']['name'] + '.json')
+        write_data_to_file(template, SVARS['base_dir'] + '/' + folder + '/templates/' + file_name + '.json')
 
 
 def process_imagestream(source, folder, location_list, imagestream):
@@ -213,20 +233,22 @@ def process_imagestream(source, folder, location_list, imagestream):
         if 'regex' in location_list:
             matches = re.search(r'(^' + location_list['regex'] + ')', stream["metadata"]["name"])
         if matches or 'regex' not in location_list:
+            file_name = check_file_name(SVARS['base_dir'] + "/" + folder + "/imagestreams/", \
+                                stream["metadata"]["name"] + ("-" + location_list["suffix"] if 'suffix' in location_list else ''), '.json')
             index_data = {
                 'name': stream['metadata']['name'],
                 'docs': location_list['docs'] if 'docs' in location_list else '',
                 'source_url': location_list['location'],
-                'path': source + '/' + folder + '/imagestreams/' + stream["metadata"]["name"] + ("-" + location_list["suffix"] if 'suffix' in location_list else '') + ".json"
+                'path': (source + '/' if source in SVARS["sources"] else '') + folder + '/imagestreams/' + file_name + ".json"
                 }
             append_to_index(source, folder, 'imagestreams', index_data)
-            write_data_to_file(stream, SVARS['base_dir'] + "/" + folder + "/imagestreams/" + stream["metadata"]["name"] + ("-" + location_list["suffix"] if 'suffix' in location_list else '') + ".json")
+            write_data_to_file(stream, SVARS['base_dir'] + "/" + folder + "/imagestreams/" + file_name + '.json')
 
 
 def create_indexes():
     """ Creates the index.json and README.md files """
 
-    for source in SVARS['sources']:
+    for source in SVARS['tags']:
         write_data_to_file(SVARS['index'][source], source + '/index.json')
 
         with open(source + '/README.md', 'a') as index_file:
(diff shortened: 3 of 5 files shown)


coreydaley:
@dinhxuanvu The readme needs to be updated with descriptions of the new fields.

dinhxuanvu:
@coreydaley Hey Corey. Will do. Just FYI, this PR will not be merged into upstream. I created a PR so code reviewers can see the commits + diff a bit easier. If the PR is approved, it may end up in a different branch all by itself.

dinhxuanvu:
@abhgupta PTAL

dinhxuanvu:
@coreydaley @luciddreamz PR is updated so it can be merged into upstream/master instead of living in a separate branch. The "--online" option is optional and can be enabled for Online-specific usage. Without using that flag, the functionality remains the same as the original code. Please review and merge. Thanks.

bparees:
This pr is not being merged here, correct?

dinhxuanvu:
@bparees This PR was supposed to be in a new branch as it was doing different thing than original code. However, I changed the PR so it can be merged into master. The "import online-specific templates/imagestreams" task is optional selection that can be enabled. Otherwise, the code is working as originally expected.

bparees:
i'm not a huge fan of this.  The intent of the openshift/library was to be a repository of all templates/imagestreams and then have various products (online, ocp, origin, etc) pull out the templates/imagestreams they cared about.  if we start segregating within this repo then i feel like we're moving backwards again.

dinhxuanvu:
I understand your point but this PR doesn't change the original functionality of the repo at all. The import_content.py still does the job of pulling templates and imagestreams into directory as expected (nothing changes on that part). Only when it is used with flag "--online", then it will pull online-specific templates/imagestreams. That flag is optional as well. The only noticeable change is the tag (openshift: online: starter/professional) and those are also optional as you don't add those either. If online team need those tags, they can make a PR and add those themselves.

bparees:
>  That flag is optional as well. The only noticeable change is the tag (openshift: online: starter/professional) and those are also optional as you don't add those either. I

right, but it's all exposed as "here's stuff you can do w/ the metadata that drives the import" and i do not want that metadata to start showing up in this repo (at least not in the master branch).

if we can all swear that we're going to not merge PRs that actually introduce those tags into the community/official yaml files, I guess it's ok, it's just a bit awkward.

dinhxuanvu:
I don't mind for the code to stay in a separate branch. However, Jake Lucky @luciddreamz really wanted the changes to be master as it is easier to maintain the community/official yaml files in master altogether than in different branches. Anyway, I'll let him explain his thoughts and reasoning a bit before we decide what to do next.

bparees:
i can live with it as long as we aren't going to use it here.

community.yaml
```diff
@@ -3,6 +3,10 @@ data:
     templates:
       - location: https://raw.githubusercontent.com/luciddreamz/laravel-ex/master/openshift/templates/laravel-mysql.json
         docs: https://github.com/luciddreamz/laravel-ex/blob/master/readme.md
+        openshift:
+          online:
+            - starter
+            - professional
```
bparees: @dinhxuanvu @luciddreamz if we can agree to remove the actual usage of the "online" tagging from this PR, i'm ok to merge it.

dinhxuanvu: It is a bit of a pickle for me given without the tags, the code change doesn't do anything. The new code relies on the tags to select the templates/imagestreams which are needed for online. I wonder if we can create a new branch named "online" for this PR to be merged into if you insist not keeping the online tags.
I'm tagging Abhishek @abhgupta for further discussion here.

bparees: well like i said, i would like for library to remain the central repository of all the imagestreams/templates that can be consumed by any user/product.

starting to split off specific templates into "this one is for online, this one is for OCP, this one is for origin, this one is for dedicated" seems like a job for the consumers of the content in this repo, not this repo itself.  (if you're Online, you can decide what set of templates you want to grab from the openshift/library.  You shouldn't have to go update the openshift/library itself).

import_content.py
```diff
@@ -262,23 +294,59 @@ def main():
...
+                                        if item_type == 'templates':
+                                            if not os.path.exists(os.path.join(SVARS['base_dir'], directory, "templates/examples")):
+                                                os.makedirs(os.path.join(SVARS['base_dir'], directory, "templates/examples"))
+                                            process_template(source, directory, item, dict_data)
+                                        elif item_type == 'imagestreams':
+                                            if not os.path.exists(os.path.join(SVARS['base_dir'], directory, "imagestreams")):
+                                                os.makedirs(os.path.join(SVARS['base_dir'], directory, "imagestreams"))
+                                            process_imagestream(source, directory, item, dict_data)
+                                else:
+                                    print "No 'openshift' tag found for " + item_type + " " + item["location"] +  ". Skipping..."
```
coreydaley: Should use the "message" function instead of print here
message(action, sub_action, text)

import_content.py
```diff
@@ -236,15 +245,38 @@ def main():
...
     args = parser.parse_args()
 
+    global online
+    # Flag online option if selected
+    online = args.online
+
     if not os.path.exists('tmp'):
         os.makedirs('tmp')
 
+    # Remove the templates + imagestreams in free and paid directory
```
coreydaley: Why not just remove the [free, paid] directories instead of removing all of the sub directories one at a time?

import_content.py
```diff
@@ -262,23 +294,59 @@ def main():
                 message('Error', 'YAML', 'Variable replacement caused invalid YAML')
                 exit(1)
         if 'data' in doc:
-            if os.path.exists(source):
-                message("Deleting", "folder", source)
-                shutil.rmtree(source)
+            if not online:
```
coreydaley: if not online and os.path.exists(source):
Shouldn't be a need for two levels of if here

import_content.py
```diff
@@ -262,23 +294,59 @@ def main():
...
+                if os.path.exists(source):
+                    message("Deleting", "folder", source)
+                    shutil.rmtree(source)
             for folder, contents in doc['data'].items():
-                if not os.path.exists(os.path.join(SVARS['base_dir'], folder)):
-                    os.makedirs(os.path.join(SVARS['base_dir'], folder))
+                if not online:
```
coreydaley: if not online and not os.path.exists(.......)

import_content.py
```diff
@@ -262,23 +294,59 @@ def main():
...
-                                process_imagestream(source, folder, item, dict_data)
-    create_indexes()
+                            # Check if online flag is true or not:
+                            if online:
+                                # Check for the "openshift" tag in YAML file
+                                if "openshift" in item:
+                                    status, dict_data = fetch_url(item['location'])
+                                    tier = item.get("openshift").get("online")
+                                    # Check for "starter" and "professional" option
+                                    if "starter" in tier:
```
coreydaley: It seems like it would make more sense for the tag and the folder to match starter/starter or free/free, kind of confusing

import_content.py
```diff
@@ -262,23 +294,59 @@ def main():
...
+                                        directory = "free/"
+                                        if item_type == 'templates':
+                                            if not os.path.exists(os.path.join(SVARS['base_dir'], directory, "templates/examples")):
+                                                os.makedirs(os.path.join(SVARS['base_dir'], directory, "templates/examples"))
+                                            process_template(source, directory, item, dict_data)
+                                        elif item_type == 'imagestreams':
+                                            if not os.path.exists(os.path.join(SVARS['base_dir'], directory, "imagestreams")):
+                                                os.makedirs(os.path.join(SVARS['base_dir'], directory, "imagestreams"))
+                                            process_imagestream(source, directory, item, dict_data)
+                                    if "professional" in tier:
```
coreydaley: same as above, tag and folder should match

import_content.py
```diff
@@ -169,15 +172,19 @@ def process_template(source, folder, location_list, template):
         if valid:
```
coreydaley: Need to do a rebase from master which will fix these couple of lines (valid isn't set anywhere)

import_content.py
```diff
@@ -262,23 +294,59 @@ def main():
...
-                            if item_type == 'templates':
-                                process_template(source, folder, item, dict_data)
-                            elif item_type == 'imagestreams':
-                                process_imagestream(source, folder, item, dict_data)
-    create_indexes()
+                            # Check if online flag is true or not:
+                            if online:
+                                # Check for the "openshift" tag in YAML file
+                                if "openshift" in item:
+                                    status, dict_data = fetch_url(item['location'])
```
coreydaley: Need to rebase to fix these couple of lines (status is not used)

community.yaml
abhgupta: I would agree with @bparees on this one. My proposal (sort of on the lines of what I had tried to share with you @dinhxuanvu ) is that we maintain a list/names of templates that we want for Online free and paid in a separate location - say, a file under free/ and one under paid/ that lists the templates and imagestream names that pertained to the individual environment. 

The jenkins job could process all the templates and image streams and then submit PR to the online repo with the changes to the templates/imagestreams that pertained to the specific environment.

luciddreamz: Just for clarity: I wanted to do the tagging for online here, but probably leave the scripts as-is. The reason for the online tagging being here -- When someone asks how they can get their template/image added to Online, we would simply point them at this repo and say "Make sure your thing is in Library and tagged for OpenShift Online."

dinhxuanvu:
@bparees The PR is ready for review again.
1. Change script to be more generic
2. Allow custom tags via cli/flags
3. Simplify tags to be simpler such as "online-starter" under "openshift"
4. File collision detection is added
5. Index and README are added to folders to allow users to locate files + sources.

import_content.py
```diff
@@ -249,15 +271,31 @@ def main():
...
 
+    # Get custom tags information and add to SVARS["tags"]
+    # Otherwise, use the default tags as community and official
+    tags = args.tags
+    if tags:
+        SVARS['tags'] = tags.split(",")
+        # Delete the custom tag(s) directori(es)
+        for tag in SVARS['tags']:
+            message("Deleting", "folder", tag)
+            shutil.rmtree(tag)
```
coreydaley: You need to check that the directory exists before removing it or it errors.  
if os.path.exists(tag):

README.md
```diff
@@ -113,6 +121,8 @@ Listings in the **official.yaml** file will be created in a sub folder of the  *
 
 The **folder_name** is a sub folder which represents a logical grouping for a set of templates or image-streams in the top level **official** or **community** folders.
 
+For "online" option, templates are imported into <free/paid>/templates/examples/ directory while imagestreams are imported into <free/paid>/imagestreams/
```
coreydaley: Shouldn't these be <online-starter/online-professional> for the directory names?

import_content.py
```diff
@@ -249,15 +271,31 @@ def main():
...
+    else:
+        SVARS['tags'] = ['official', 'community']
```
coreydaley: It would be better if this set SVARS['tags'] to the contents of SVARS['sources'] so it does not need to be updated in two places.

import_content.py
```diff
@@ -279,21 +317,45 @@ def main():
...
-                                exit(1)
-                            if item_type == 'templates':
-                                process_template(source, folder, item, dict_data)
-                            elif item_type == 'imagestreams':
-                                process_imagestream(source, folder, item, dict_data)
+                            # Check if custom tags are provided:
+                            if tags:
+                                # Check for the "openshift" tag in YAML file
+                                if "openshift" in item:
+                                    for tag in SVARS["tags"]:
```
coreydaley: Couldn't you just iterate over the contents of openshift and create the directories that are needed?

import_content.py
```diff
@@ -279,21 +317,45 @@ def main():
...
+                                        if not os.path.exists(os.path.join(SVARS['base_dir'], tag)):
+                                            os.makedirs(os.path.join(SVARS['base_dir'], tag))
+                                        status, dict_data = fetch_url(item['location'])
+                                        if status != 200:
+                                            message('Error', 'Not Found', item['location'])
+                                            exit(1)
+                                        options = item.get("openshift")
+                                        # Import imagestreams/templates for the option
+                                        if tag in options:
```
coreydaley: You could also eliminate this check by making the above change

README.md
```diff
@@ -53,14 +53,14 @@ Community templates and image streams are **not** provided or supported by Red H
 ### Running the Script
 
     $ make import
```
coreydaley: You may want to add a make onlineimport that runs the import_content.py -t online-starter,online-professional

import_content.py
```diff
@@ -162,6 +162,25 @@ def write_data_to_file(data, path):
     target_file.write(json.dumps(data, sort_keys=True, indent=4, separators=(',', ': ')))
     target_file.close()
 
+def check_file_name(path, file_name, extension):
```
coreydaley: Have you run into filename collisions?

dinhxuanvu: No I haven't but given with custom tags, we do put templates/imagestreams all into their respective directories so there is possibility of filename collisions. Ben insists I have a check for it. It is just to be careful. That is all.

import_content.py
dinhxuanvu: Will address this. Thanks

README.md
dinhxuanvu: Will fix this as well. Thanks.

README.md
dinhxuanvu: We discussed this so no changes are needed.

import_content.py
dinhxuanvu: As we discussed on IRC, the tags in yaml files are custom and can be expanded in the future. They are not limited to just online-starter/professional. Users can choose to import only one tag or multiple tags via CLI so one way or the other, the code needs to compare the list of tags provided in CLI and the tags are listed under "openshift" in yaml file because they are not always the same.

import_content.py
dinhxuanvu: Same as above.

README.md
```diff
@@ -89,11 +89,19 @@ That's it!  Your pull request will be reviewed by a member of the OpenShift Team
             regex: # (optional) matched against ['metadata']['name'] in the json file
             suffix: # (optional) suffix for the file that is created ex: ruby-<suffix>.json
             docs: # (optional) web address of the documentation for this image-stream
+            # Custom tags for OpenShift are added below if needed
+            openshift: # (optional) flag for OpenShift
```
bparees: why name this "openshift"?  they are just optional tags, right?  how about calling it "tags"?

dinhxuanvu: It is simply a name that Jake used when he created the tag schema. It basically indicates the product name. We can change it to whatever we want as long as we all agree to it.

Regarding the folder structure, the multi-layer sub-directories are unnecessary for Online and for another custom tags from what I see. People run the script to import selected templates/IS based on tags because they want to wrap quickly and all in the same place. With the name of the framework/languages are already on the name of the files, why we have to divide them in to sub-folder of framework/lang and then another sub-folder for templates or IS? That may be helpful when you use the script with original functionality (AKA wrap all of them). With custom tags, the amount of templates/IS are imported should be way less than the total available ones in yaml files.

dinhxuanvu: Also, even if we go back to the same directory structure, there is a if statement code from 321 to 324 needs to be removed and some minor changes but the overall code is the same. So not sure if that simplifies much.

bparees: all the code to handle filename collisions would go away (including the helper function that renames them).

and all the logic that sets a different directory based on whether this is a tags run or not would go away (you could just set the target directory once at the beginning.  having a target dir for each tag is unnecessary, you can get the same behavior by just running the script once for each tag, with a different target dir.  Also the existing behavior precludes importing multiple tags into a single dir, which could be more useful).

the only logic that would need to stay is "does this item match one of the user provided tags", rather than the current dual codepaths of "are we in "import via tag" mode or "are we in normal import mode".


> . With the name of the framework/languages are already on the name of the files, why we have to divide them in to sub-folder of framework/lang and then another sub-folder for templates or IS?

those filenames are by convention only, nothing enforces them.  There's nothing stopping a user from creating a template named "mytemplate" and adding it to the library.  It should not be counted on to provide any uniqueness, meaning, or identity.

As for sub-folders for imagestreams vs templates, that's just a good logical division of content types.

bparees: So i know we agreed to this approach during the call, and if you are standing by the flat filesystem as a strong requirement, we can leave it, but i really think we've overly complicated this code to solve what i'm not convinced is a valid use case requirement.

dinhxuanvu: 1. First of all, there are sub-folder for templates and imagestreams (IS) to separate the templates and IS files. The only difference is that all templates are under the same <tag>/templates directory and same for IS. There is no the middle part AKA framework/language sub-folder such as <tag>/<framework>/templates. So it is a flatter and simpler structure compared to what we have originally.

2. Even though the file collision situation hasn't happened before, it is possible for that situation happen even with the original code. The name of the file is formed using the "name" field in "metadata" from the JSON file itself. So technically, if there are two templates with the same name in the "name" fields and happen to be in the same group of framework/IS and in the same community or official yaml then the file collision will happen. For example, look at the cakephp-mysql-example/persistent.json in community YAML, if the "name" fields are both "cakephp-mysql-example" in JSON files, then the file collision will happen and you probably won't even notice because the file is just overwritten.

3. I have thought about using the target directories for tags (hence the list of tags) as a 'for' loop on top of the 'for' loop for "source". It will work well with the tags. However, it won't work with the original functionality given there are no community/official tags. The way the code is working originally, is basically looping through the 2 YAML files and create community/official folder as it goes and add templates/IS into the folder.

dinhxuanvu:
Will be replaced by #30.

PR #19, opened 2017-05-08
Fix the incorrect nodejs imagestream URL for NodeJS Centos7

dinhxuanvu:
The imagestream file in bucharest-gold/origin-s2i-nodejs repo is
recently renamed from image-stream.json to image-streams.centos7.json.
This PR will address the issue by refering to the correct file.

Signed-off-by: Vu Dinh <vdinh@redhat.com>


dinhxuanvu:
@coreydaley @bdecoste @luciddreamz PTAL

coreydaley:
@dinhxuanvu The filename that is produces (centos7-s2i-nodejs-bucharest-gold.json) doesn't seem consistent with the rest of the filenames.

Can we get it changed to something like s2i-nodejs-bucharest-gold-centos7.json ?

bparees:
@dinhxuanvu please run import_content.py and include the new/changed files w/ your PR so we can review the changes that this will actually introduce.

dinhxuanvu:
@coreydaley I have no control over the upstream repo so you need to bring it up to owner of that repo to address the issue with the name.
@bparees Ran the script as you requested.

PR #27, opened 2017-06-14
add httpd imagestream and template to the library

bparees:


bparees:
dependent on https://github.com/openshift/origin/pull/14660 before import can be run.

bparees:
@coreydaley ptal

bparees:
@sspeiche fyi

official.yaml
```diff
@@ -68,6 +68,16 @@ data:
         docs: https://github.com/openshift/cakephp-ex/blob/master/README.md
       - location: https://raw.githubusercontent.com/openshift/cakephp-ex/master/openshift/templates/cakephp-mysql-persistent.json
         docs: https://github.com/openshift/cakephp-ex/blob/master/README.md
+  httpd:
+    imagestreams:
+      - location: https://raw.githubusercontent.com/openshift/origin/master/examples/image-streams/image-streams-rhel7.json
+        docs: https://github.com/sclorg/httpd-container/blob/master/README.md
+        regex: httpd
+        suffix: rhel7
+  httpd:
```
coreydaley: This httpd->templates should be combined with the above httpd just like nodejs (https://github.com/openshift/library/blob/master/official.yaml#L30-L42), there is no reason to have two separate top level httpd entries

bparees:
@coreydaley fixed

coreydaley:
@bparees lgtm

PR #28, opened 2017-06-19
Infinispan persistent

slaskawi:
https://issues.jboss.org/browse/ISPN-7938

### DIFF
--- a/community/infinispan/imagestreams/infinispan-centos7.json
+++ b/community/infinispan/imagestreams/infinispan-centos7.json
@@ -18,22 +18,9 @@
                 },
                 "from": {
                     "kind": "ImageStreamTag",
-                    "name": "9.1"
+                    "name": "9.0"
                 },
                 "name": "latest"
-            },
-            {
-                "annotations": {
-                    "description": "Infinispan 9, a high performance data grid solution",
-                    "iconClass": "icon-infinispan",
-                    "openshift.io/display-name": "Infinispan 9.1",
-                    "tags": "java,database,datagrid,jboss"
-                },
-                "from": {
-                    "kind": "DockerImage",
-                    "name": "jboss/infinispan-server:9.1.0.Alpha1"
-                },
-                "name": "9.1"
             },
             {
                 "annotations": {
--- a/community/infinispan/templates/infinispan-persistent.json
+++ b/community/infinispan/templates/infinispan-persistent.json
@@ -1,0 +1,308 @@
+{
+    "apiVersion": "v1",
+    "kind": "Template",
+    "labels": {
+        "template": "infinispan-persistent"
+    },
+    "metadata": {
+        "annotations": {
+            "description": "Infinispan 9 (Persistent)",
+            "iconClass": "icon-infinispan",
+            "tags": "java,database,datagrid,jboss",
+            "template.openshift.io/documentation-url": "http://infinispan.org/documentation/",
+            "template.openshift.io/long-description": "Infinispan is a high performance, scalable, key/value data grid solution.",
+            "template.openshift.io/provider-display-name": "Red Hat, Inc.",
+            "template.openshift.io/support-url": "http://infinispan.org"
+        },
+        "name": "infinispan-persistent"
+    },
+    "objects": [
+        {
+            "apiVersion": "v1",
+            "kind": "ServiceAccount",
+            "metadata": {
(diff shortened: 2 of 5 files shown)


coreydaley:
@bparees lgtm

community/infinispan/imagestreams/infinispan-centos7.json
```diff
@@ -11,21 +11,34 @@
...
+                    "kind": "DockerImage",
+                    "name": "jboss/infinispan-server:9.1.0.Alpha1"
+                },
+                "name": "9.1"
+            },
+            {
+                "annotations": {
+                    "description": "Infinispan 9, a high performance data grid solution",
+                    "iconClass": "icon-infinispan",
                     "openshift.io/display-name": "Infinispan 9 (Latest)",
```
bparees: this shouldn't say latest anymore, right?

community/infinispan/imagestreams/infinispan-centos7.json
```diff
@@ -11,21 +11,34 @@
         "tags": [
             {
                 "annotations": {
-                    "description": "Infinispan 9, a high performance data grid solution",
+                    "description": "Infinispan 9.1, a high performance data grid solution",
                     "iconClass": "icon-infinispan",
-                    "openshift.io/display-name": "Infinispan 9 (Latest)",
+                    "openshift.io/display-name": "Infinispan 9",
```
bparees: The way we do this for our other images, the "latest" tag we just describe as "XXX Latest", we don't indicate the version.  Then you don't have to update the description when you point the latest tag to a newer version, you just change what tag it points to:

https://github.com/openshift/origin/blob/master/examples/image-streams/image-streams-centos7.json#L63-L75

We also don't refer to the "current" latest version as "latest', but that's up to you:
https://github.com/openshift/origin/blob/master/examples/image-streams/image-streams-centos7.json#L110-L123

bparees: also in general you probably want to warn people away from using the "latest" tag since it makes them susceptible to picking up an incompatible version change (e.g. they will pick up infinispan 10 automatically when this tag gets updated to point there, that could break them).  This is what we put in our "latest" tag descriptions:
`WARNING: By selecting this tag, your application will automatically update to use the latest version of Ruby available on OpenShift, including major versions updates.`

slaskawi: Got it! Thanks for clarification. I'll apply both suggestions than.

BTW, could you please tell me how I can supply custom icon? Currently I'm using `"iconClass": "icon-infinispan"` but I have no idea where to put the png file.

community/infinispan/imagestreams/infinispan-centos7.json
slaskawi: Good eye! Thanks Ben!

community/infinispan/imagestreams/infinispan-centos7.json
bparees: the options for icons are described here:
https://docs.openshift.org/latest/dev_guide/templates.html#writing-description

see bullet 5 under the example.  If nothing in those lists is suitable then you will have to talk to @jwforres about the process for adding new icons.  (also if icon-infinispan isn't in that list today, then your template isn't going to display the right icon)

slaskawi: Exactly. All I can see is some default icon. Thanks for the hint.

slaskawi: I sent a query to Jessica. In the mean time, all comments addressed and ready for review.

community/infinispan/templates/infinispan-persistent.json
```diff
@@ -0,0 +1,308 @@
+{
+    "apiVersion": "v1",
+    "kind": "Template",
+    "labels": {
+        "template": "infinispan-persistent"
+    },
+    "metadata": {
+        "annotations": {
+            "description": "Infinispan 9.1 (Persistent)",
```
bparees: are you going to update your ephemeral template to 9.1 also?

slaskawi: Sure, good idea. 

BTW, after small discussion with @tristantarrant we decided to remove `9.1.0.Alpha1` image. We will be shipping only finals.

slaskawi:
@bparees Updated.

PR #30, opened 2017-06-19
Add tag support on import

bparees:


bparees:
@dinhxuanvu this is how simple this change can be if we don't worry about using a different directory structure when importing with a tag filter.

@coreydaley ptal

coreydaley:
@bparees This is close, but there are a few places (like create_indexes()) that rely on there being content in SVARS['index'][source] that also need updated.

bparees:
I don't think the online team even needs the index or readme content so we could just skip generating those if tags are specified, but sure we can also just add the same filtering logic in create_indexes().

dinhxuanvu:
1. Ben is correct as my team doesn't care about index/README files but if we plan to introduce new tags in the future for other teams, they may need those. For now, it is fine.

2. This is technically working in term of of importing the correct items with tags. However, every item ends up in the same community and official folders, there are no way to tell which items belong to online-starter/online-professional if I select both tags in the CLI. The list of items are the same for both tags right now but they may be different in the future. So it is working well if I run the script twice and get the items I need for each tag separately.

bparees:
>  So it is working well if I run the script twice and get the items I need for each tag separately.

I would argue that aligns w/ general unix principles.  I could see an argument for taking a "target directory" as an arg (then you'd still have to run it twice, but you wouldn't clobber your results on the second run) but I don't think the complexity of adding logic to "extract the content into a directory that matches the tag name" is justified given how easy it is to run the tool, tar up what you got, run the tool again(with a different tag), tar up what you got.

bparees:
@coreydaley i've added the index filtering, ptal.

dinhxuanvu:
I'm fine with this solution as generic and simple way to just get items with the tags.

import_content.py
```diff
@@ -223,7 +224,7 @@ def process_imagestream(source, folder, location_list, imagestream):
             write_data_to_file(stream, SVARS['base_dir'] + "/" + folder + "/imagestreams/" + stream["metadata"]["name"] + ("-" + location_list["suffix"] if 'suffix' in location_list else '') + ".json")
 
 
-def create_indexes():
+def create_indexes(filter_tags):
     """ Creates the index.json and README.md files """
 
     for source in SVARS['sources']:
```
coreydaley: The rest of the method after this line needs to be wrapped in

`if source in SVARS['index']:`

import_content.py
```diff
@@ -235,25 +236,45 @@ def create_indexes():
                 for section, items in sorted(sections.items()):
                     index_file.write('## ' + section + '\n')
                     for item in items:
+                        if not has_tag(item,filter_tags):
```
coreydaley: nit
item, filter_tags

import_content.py
```diff
@@ -235,25 +236,45 @@ def create_indexes():
...
+                            continue
                         index_file.write('### ' + item['name'] + '\n')
                         index_file.write('Source URL: [' + item['source_url'] + '](' + item['source_url'] + ' )  \n')
                         if 'docs' in item and item['docs'] != '':
                             index_file.write('Docs: [' + item['docs'] + '](' + item['docs'] + ')  \n')
                         index_file.write('Path: ' + item['path'] + '  \n')
 
 
+def has_tag(item,filter_tags):
```
coreydaley: nit
item, filter_tags

import_content.py
```diff
@@ -235,25 +236,45 @@ def create_indexes():
...
     parser = argparse.ArgumentParser(description='Build OpenShift template and image-stream library')
+    parser.add_argument("-t", "--tags", nargs='?', help="Select specific tag(s) to import templates/imagestreams (separated by comma ',')")
     args = parser.parse_args()
 
     if not os.path.exists('tmp'):
         os.makedirs('tmp')
 
+    # Get custom tags information and add to SVARS["tags"]
+    # Otherwise, use the default tags as community and official
+    tags = args.tags
```
coreydaley: Either update the comment, or change this to be SVARS["tags"] = args.tags and update the rest of the code to reference SVARS["tags"], which would allow you to change your has_tag function to just take an item and split the SVARS["tags"], same with create_indexes.

import_content.py
```diff
@@ -279,11 +300,11 @@ def main():
                 message("Deleting", "folder", source)
                 shutil.rmtree(source)
             for folder, contents in doc['data'].items():
-                if not os.path.exists(os.path.join(SVARS['base_dir'], folder)):
-                    os.makedirs(os.path.join(SVARS['base_dir'], folder))
                 for item_type in ['imagestreams', 'templates']:
                     if item_type in contents and len(contents[item_type]) > 0:
                         for item in contents[item_type]:
+                            if not has_tag(item,tags):
```
coreydaley: nit
item, tags

import_content.py
bparees: only if you hate nil pointer errors......

import_content.py
bparees: i'll update the comment. not a fan of global map env variables :)

bparees:
@coreydaley updated.  note that i did not actually need the filter in create_index so i've removed that.  The content driving the index is only stuff that already made it through the filter.

PR #33, opened 2017-07-01
fix renamed apicast quickstart

bparees:


bparees:
@coreydaley fyi

bparees:
(it also seems backwards that we are pulling this quickstart from the openshift ansible installer repo.  that certainly isn't an official location... any idea where this one came from?)

PR #38, opened 2017-09-07
Adds Apache HTTP Server to OpenShift Online

luciddreamz:


luciddreamz:
@bparees Note that my only actual change was adding tags to official.yaml for httpd to get added to Online.

bparees:
lgtm

PR #42, opened 2017-11-10
WIP: bump xpaas_version to ose-v1.4.7

rcernich:
Signed-off-by: rcernich <rcernich@redhat.com>

v1.4.7 should release next week


bparees:
can you run the import script("make import") and include the updates in this PR?

bparees:
@dinhxuanvu fyi/ptal too

rcernich:
ok, not sure if i did the right thing or not.  hopefully, ``make import`` was all that was required.

dinhxuanvu:
I thought this is gonna be an upgrade from 1.4.x to 1.4.7 but some of templates is downgrading from 1.5.0 to 1.4.7. Is that right? @rcernich

bparees:
@rcernich well it's passing travis now, so you probably did it right :)

bparees:
not sure why we ever had "1.5.0" stuff in here, there's never been a 1.5.0 xpaas release.

rcernich:
There's no reversion.  The version update was just to make it consistent with the tag in application-templates.  The previous should have been 1.4.5 instead of 1.5.0.  Thanks for noticing. ;)

bparees:
thanks @rcernich !

PR #43, opened 2017-11-10
Revert "WIP: bump xpaas_version to ose-v1.4.7"

bparees:
Reverts openshift/library#42

(images aren't actually available yet)


bparees:
@dinhxuanvu @rcernich fyi.

PR #51, opened 2018-01-24
Removing starter tag from 3scale template

maneta:
As for today, the 3scale resources consumption is not suitable for starter users.

We are working on lowering the resources consumption, but meanwhile we have decided to remove it from the starter templates.

`make import` and `make verify` has been executed:

```bash
Diffing current official directory against freshly generated official directory
SUCCESS: Generated official directory up to date.
Diffing current community directory against freshly generated community directory
SUCCESS: Generated community directory up to date.
```

Thanks!


bparees:
lgtm

PR #56, opened 2018-02-13
get jenkins content from jenkins repo and add nginx imagestreams

bparees:


bparees:
@coreydaley ptal

official.yaml
```diff
@@ -176,6 +176,11 @@ data:
           - online-starter
           - online-professional
   nginx:
+    imagestreams:
+      - location: https://raw.githubusercontent.com/openshift/origin/master/examples/image-streams/image-streams-rhel7.json
+        docs: https://github.com/sclorg/nginx-container/blob/master/README.md
+        regex: nginx
+        suffix: rhel7
     templates:
```
coreydaley: Should these nginx templates be in the community.yaml also?

official.yaml
```diff
@@ -100,15 +100,15 @@ data:
           - online-professional
   jenkins:
     templates:
```
coreydaley: Should these jenkins templates be in the community.yaml also?

official.yaml
bparees: I don't think so.  I'm not sure why we have the other templates in both locations but it doesn't make sense to me to have the same template (and they are the same) as both an "official" and "community" template.  If anything i think we should remove the community templates for the other ones.

official.yaml
bparees: same answer

official.yaml
coreydaley: I would say "for ease of use for the users".  Users are either going to load the rhel (official) or centos (community) yaml files.  In this case, if they loaded the centos one, they would be missing the associated templates.

bparees: not sure what you mean by "load the xxx yaml file".  

they can't do anything meaningful w/ the yaml file itself.

openshift/library is just a repository of content, you pick and choose what you want out of it.

coreydaley: I mean to say if they write a utility to get and parse the xxx.yaml file, they will be missing the templates.  I defer to your judgment on that.

bparees: having it in both places implies there might be a difference between them.  (plus it just bloats the repo).  it would also effectively just make "community" be a superset of "official".  so yeah, i'm gonna stick w/ not putting them in both places.

PR #59, opened 2018-02-27
update to import from sclorg repos instead of origin

bparees:
fixes https://github.com/openshift/library/issues/62

### DIFF
--- a/community.yaml
+++ b/community.yaml
@@ -26,7 +26,7 @@ data:
         suffix: ubuntu14
   wildfly:
     imagestreams:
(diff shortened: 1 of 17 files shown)


bparees:
ok this should be good to go as soon as i get confirmation the postgres images w/ the check-container script are all published (it looks like they are).

jorgemoralespou:
@bparees I thin wildfly's Imagestreams have been updated by you to add a .0 to 11 and 12, so this is already old :-(

PR #60, opened 2018-03-02
Add tags to FIS templates & imagestreams

jamesnetherton:


bparees:
@sspeiche @rettori

sspeiche:
+1 This fine. This was actually missed from the original PR.

sspeiche:
@dinhxuanvu @abhgupta Can we get this added to Pro sooner than later? We need to add the statement like we have in other templates

PR #64, opened 2018-03-13
remove bucharest gold, imagestream is broken

bparees:

PR #65, opened 2018-03-13
src: restore bucharest-gold image streams

lance:
This commit reverts https://github.com/openshift/library/pull/64 (e9e8f13438a29bd594945df1755e6f74646381e0) and updates the name of the image stream to `image-stream.yml`.


bparees:
thanks @lance

PR #68, opened 2018-03-29
Add .NET Core templates

tmds:
@bparees PTAL

Will this sync automatically with openshift/origin?

### DIFF
--- a/official.yaml
+++ b/official.yaml
@@ -567,6 +567,13 @@ data:
         tags:
           - online-professional
   dotnet:
+    templates:
+      - location: https://raw.githubusercontent.com/redhat-developer/s2i-dotnetcore/master/templates/dotnet-example.json
+        docs: https://github.com/redhat-developer/s2i-dotnetcore/blob/master/README.md
+      - location: https://raw.githubusercontent.com/redhat-developer/s2i-dotnetcore/master/templates/dotnet-pgsql-persistent.json
+        docs: https://github.com/redhat-developer/s2i-dotnetcore/blob/master/README.md
+      - location: https://raw.githubusercontent.com/redhat-developer/s2i-dotnetcore/master/templates/dotnet-runtime-example.json
+        docs: https://github.com/redhat-developer/s2i-dotnetcore/blob/master/README.md
     imagestreams:
       - location: https://raw.githubusercontent.com/redhat-developer/s2i-dotnetcore/master/dotnet_imagestreams.json
         docs: https://github.com/redhat-developer/s2i-dotnetcore/blob/master/README.md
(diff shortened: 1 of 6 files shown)


bparees:
> Will this sync automatically with openshift/origin?

no, there's no automatica syncing from library to origin.  We manage that semi-manually.

But in this case, we don't ship/carry any .net templates in origin today.  (We do carry the imagestream).  Given that these are community templates, I wouldn't expect us to, i'd expect people who want them to get them from here.

community.yaml
```diff
@@ -191,6 +191,13 @@ data:
         regex: redis
         suffix: centos7
   dotnet:
+    templates:
+      - location: https://raw.githubusercontent.com/redhat-developer/s2i-dotnetcore/master/templates/dotnet-example.json
```
bparees: we don't put content in both places, so are these supported(official) or community?

bparees:
@tmds looks like you didn't add the imported files to your commit

community.yaml
tmds: supported, we update these templates and verify they work for each release

bparees: ok, then i'd only add them to the official.yaml

tmds:
@bparees removed community, added imported files

tmds:
@bparees it seems nodejs has same templates both in official and community?

https://github.com/openshift/library/blob/a51af1ccfa5d5b6feeb45e191852acd6b8e7603e/official.yaml#L43-L52

https://github.com/openshift/library/blob/a51af1ccfa5d5b6feeb45e191852acd6b8e7603e/community.yaml#L73-L80

bparees:
> @bparees it seems nodejs has same templates both in official and community?

thanks, looks like we have a few to clean up.

tmds:
ok, will these templates find their way to openshift/origin as part of the semi-manual syncing?

tmds:
There are a number of quickstarts here: https://github.com/openshift/origin/tree/master/examples/quickstarts, which made me think .NET Core templates should be there too.

tmds:
> It would probably be best to just leverage service provisioning/binding.

I don't know what that means.

> I'd like to get rid of templates that just hook 2 things together.

Ah, then we shouldn't include these.

> Plus a -1 on templates like https://raw.githubusercontent.com/redhat-developer/s2i-dotnetcore/master/templates/dotnet-runtime-example.json
as it uses Docker build strategy, this won't work online.

+1 on catching this. I'll do a PR to remove it.

tmds:
> Here's some high level info and links to the service https://blog.openshift.com/whats-new-in-openshift-3-7-service-catalog-and-brokers/

@sspeiche I'm giving this a try. One thing I notice is I don't get an overview of 'environment variables' to change the build process. In contrast, when using the template, the UI gives a list of the template parameters, which correspond to s2i environment variables for the simple templates. So the UX for the simple template is nicer than for the imagestream.

> And then we'll also have to get it into openshift-ansible for it to ship w/ OCP.

@bparees They are already in OCP. I'll do a PR to add them to origin as well.

In general, when we make changes to templates/imagestreams in s2i-dotnetcore, should we make PRs to update openshift-ansible/origin too?

bparees:
> In general, when we make changes to templates/imagestreams in s2i-dotnetcore, should we make PRs to update openshift-ansible/origin too?

yeah unfortunately that is the process currently.

PR #69, opened 2018-03-29
remove duplicated templates from community

bparees:


coreydaley:
@bparees Just removing templates that are duplicated in official? or ?

bparees:
@stevekuznetsov i don't think we have a prow merge job for this repo anyway, do we?

PR #70, opened 2018-03-30
Remove .NET Core runtime template

tmds:
This template uses the Docker build strategy, which doesn't work online.

CC @bparees


tmds:
Thanks!

PR #78, opened 2018-04-25
meta: change nodejs runtime to use rhoar-nodejs

lance:
This removes a couple of the example templates. These templates could be modified if necessary, and restored.

I have validated both the image stream and the template using openshift online starter.

### DIFF
--- a/official.yaml
+++ b/official.yaml
@@ -43,18 +43,14 @@ data:
   nodejs:
     templates:
       - location: https://raw.githubusercontent.com/sclorg/nodejs-ex/master/openshift/templates/nodejs.json
-        docs: https://github.com/sclorg/nodejs-ex/blob/master/README.md
-      - location: https://raw.githubusercontent.com/sclorg/nodejs-ex/master/openshift/templates/nodejs-mongodb.json
-        docs: https://github.com/sclorg/nodejs-ex/blob/master/README.md
-      - location: https://raw.githubusercontent.com/sclorg/nodejs-ex/master/openshift/templates/nodejs-mongodb-persistent.json
-        docs: https://github.com/sclorg/nodejs-ex/blob/master/README.md
+        docs: https://appdev.openshift.io/docs/nodejs-runtime.html
         tags:
           - online-starter
           - online-professional
     imagestreams:
-      - location: https://raw.githubusercontent.com/sclorg/s2i-nodejs-container/master/imagestreams/nodejs-rhel7.json
-        docs: https://github.com/sclorg/s2i-nodejs-container/blob/master/README.md
-        regex: nodejs
+      - location: https://raw.githubusercontent.com/bucharest-gold/centos7-s2i-nodejs/master/imagestreams/nodejs-rhel7.yml
+        docs: https://github.com/bucharest-gold/centos7-s2i-nodejs/blob/master/README.md
+        regex: rhoar-nodejs
         suffix: rhel7
         tags:
           - online-starter
--- a/community.yaml
+++ b/community.yaml
@@ -62,7 +62,7 @@ data:
         docs: https://github.com/sclorg/s2i-nodejs-container/blob/master/README.md
         regex: nodejs
         suffix: centos7
-      - location: https://raw.githubusercontent.com/bucharest-gold/centos7-s2i-nodejs/master/image-stream.yml
+      - location: https://raw.githubusercontent.com/bucharest-gold/centos7-s2i-nodejs/master/imagestreams/nodejs-centos7.yml
         docs: https://github.com/bucharest-gold/centos7-s2i-nodejs/blob/master/README.md
         suffix: bucharest-gold
   perl:


official.yaml
```diff
@@ -43,18 +43,14 @@ data:
...
+        docs: https://appdev.openshift.io/docs/nodejs-runtime.html
         tags:
           - online-starter
           - online-professional
     imagestreams:
-      - location: https://raw.githubusercontent.com/sclorg/s2i-nodejs-container/master/imagestreams/nodejs-rhel7.json
-        docs: https://github.com/sclorg/s2i-nodejs-container/blob/master/README.md
-        regex: nodejs
+      - location: https://raw.githubusercontent.com/bucharest-gold/centos7-s2i-nodejs/master/imagestreams/nodejs-rhel7.yml
+        docs: https://github.com/bucharest-gold/centos7-s2i-nodejs/blob/master/README.md
```
bparees: why does this have centos in the path if it's rhel?

official.yaml
```diff
@@ -43,18 +43,14 @@ data:
...
-      - location: https://raw.githubusercontent.com/sclorg/nodejs-ex/master/openshift/templates/nodejs-mongodb.json
-        docs: https://github.com/sclorg/nodejs-ex/blob/master/README.md
-      - location: https://raw.githubusercontent.com/sclorg/nodejs-ex/master/openshift/templates/nodejs-mongodb-persistent.json
-        docs: https://github.com/sclorg/nodejs-ex/blob/master/README.md
+        docs: https://appdev.openshift.io/docs/nodejs-runtime.html
         tags:
           - online-starter
           - online-professional
     imagestreams:
-      - location: https://raw.githubusercontent.com/sclorg/s2i-nodejs-container/master/imagestreams/nodejs-rhel7.json
```
bparees: we can't remove these unless you're providing equivalent/compatible imagestreamtags with the same name as what this provided, and you don't appear to be doing so.

official.yaml
```diff
@@ -43,18 +43,14 @@ data:
   nodejs:
     templates:
       - location: https://raw.githubusercontent.com/sclorg/nodejs-ex/master/openshift/templates/nodejs.json
-        docs: https://github.com/sclorg/nodejs-ex/blob/master/README.md
-      - location: https://raw.githubusercontent.com/sclorg/nodejs-ex/master/openshift/templates/nodejs-mongodb.json
-        docs: https://github.com/sclorg/nodejs-ex/blob/master/README.md
-      - location: https://raw.githubusercontent.com/sclorg/nodejs-ex/master/openshift/templates/nodejs-mongodb-persistent.json
-        docs: https://github.com/sclorg/nodejs-ex/blob/master/README.md
```
bparees: these are templates we ship with OCP, i don't think we can be removing them.

bparees:
@lance @sspeiche I think we need to talk about what the goal is here, but i don't think we can do it as is.

I would suggest starting by introducing this new content as an addition to the existing content, not removing the existing content that we have a bunch of dependencies on/users of.

sspeiche:
@bparees  Agree, that is what I thought we agreed to and could use the hidden tag in the online repo (or just change the display name to say "RHSCL Node.js"

bparees:
also if we're intending to replace the scl nodejs image w/ this one, i'd prefer to stop maintaining the scl images and update the imagestream to point to this new image.  (assuming it is 100% compatible w/ what it is replacing).

lance:
@bparees I've been trying to determine what the goal is now for a while. :) I suppose I was mistaken. But I wanted to at least get a PR in place so that the discussion can happen less in the abstract and more in the concrete.

@sspeiche apologies - I must have misunderstood the conversation. I am unaware of how to use a hidden tag. Can you please point me in the right direction? I saw nothing about that in the document I was referencing to create this PR.

official.yaml
lance: Is it possible to simply add a second imagestreams blog to the file here: https://github.com/openshift/library/blob/master/official.yaml#L55-L61. That was my original intent, but although the property is an array I did not see any others that were doing this. And I know for certain that some changes will need to be made to the existing templates if we just reuse those. I can make those changes and submit them as a pull request to the SCL org, or I can duplicate/modifyasnecessary those templates and put them in the bucharest-gold org as I did with the basic one.

official.yaml
lance: That's the name of the upstream repository. The product repo for these artifacts is internal. I would rather not change the name of the repository since we already have a lot of references to this all over the place. If it makes more sense, I can create a new repository with the letters 'rhel' in it that just houses the imagestream and template files. Does it really matter what the URL is?

lance:
Should I run `make import` and commit those results to this branch in order to fix CI?

official.yaml
lance: If the templates are specific to an imagestream, and we want to have more than one node.js imagestream (scl and rhoar), how does that work?

official.yaml
bparees: no, it just looked strange.

btw github does a pretty good job w/ redirects if you do choose to rename.

official.yaml
bparees: yes, you can do that.

i don't follow your point about the templates.

official.yaml
bparees: the templates will use the imagestream they reference.

users can still choose to use any imagestream directly.

lance:
Closing this for now. I believe I jumped the gun.

PR #80, opened 2018-04-25
add rhoar nodejs imagestream

bparees:


bparees:
@lance fyi this is what i would expect to see in the future.

Note that there are no changes to the templates, they will continue to use the scl nodejs imagestream named "nodejs".  Changing that is something we would need to discuss separately.

this also does not tag your new "rhoar-nodejs" imagestream into starter/pro.  Again, that would be something to discuss, the current concern is that we'd then have a "nodejs" and "rhoar-nodejs" imagestream in starter/pro.

bparees:
i merged this because the existing official.yaml and community.yaml are now pointing to broken imagestream links, so that needed to be fixed.

PR #92, opened 2018-06-07
nodejs: combine scl and rhoar community imagestreams

lance:
This commit modifies the `community.yaml` file for `nodejs`. Both the SCL images and the RHOAR images are now all specified in the new imagestream. /cc @sspeiche


community/nodejs/imagestreams/nodejs-centos7.json
```diff
@@ -1,112 +0,0 @@
-{
```
bparees: i wouldn't expect this file to be deleted...did your import work correctly?

PR #93, opened 2018-06-11
CLOUD-2593 - Deprecate AMQ 6.2

gaohoward:
Signed-off-by: Howard Gao <howard.gao@gmail.com>


bparees:
we generally do not remove imagestreamtags because people who have references to them will need those references to continue to work.  The correct way to deprecate content is to mark it hidden so it can be used, but won't show up in the console/etc by default.  For example: 
https://github.com/openshift/library/blob/master/official/nodejs/imagestreams/nodejs-rhel7.json#L39

Removing the templates is ok.

gaohoward:
@bparees Thanks, I'll mark it hidden.

gaohoward:
@bparees Hi, I restored the json file under imagestreams folder. I found that the file already has 'hidden' label, so I assume that's ok. Please review. Thanks

official.yaml
```diff
@@ -247,10 +239,6 @@ data:
         tags:
           - online-professional
     imagestreams:
-      - location: https://raw.githubusercontent.com/jboss-openshift/application-templates/{amq_version}/amq/amq62-image-stream.json
-        docs: https://access.redhat.com/documentation/en/red-hat-jboss-middleware-for-openshift/3/paged/red-hat-jboss-a-mq-for-openshift/
-        regex: jboss-amq
-        suffix: rhel7
```
bparees: you need to restore this, right?

gaohoward: @bparees Thank you. I was just wondering why it failed ci build. :)

gaohoward:
@bparees I restored the official.yaml. Thanks!

gaohoward:
@bparees Thanks! It's done now.

PR #95, opened 2018-06-21
update for latest Java S2I image and templates

rcernich:
openjdk18-openshift:1.4 should be available in the RHCC early next week.

Signed-off-by: rcernich <rcernich@redhat.com>

### DIFF
--- a/official.yaml
+++ b/official.yaml
@@ -1,5 +1,6 @@
 variables:
   xpaas_version: ose-v1.4.8-1
+  openjdk_version: ose-v1.4.14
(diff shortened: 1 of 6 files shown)


bparees:
> openjdk18-openshift:1.4 should be available in the RHCC early next week.

@rcernich please re-tag me when the images are available, I don't want to risk this getting into the wild before they are in place.  Otherwise lgtm.

rcernich:
Will do.

rcernich:
The updated jdk image has been pushed to rhcc, so this can be merged.  @bparees

PR #96, opened 2018-06-22
CLOUD-2683 Add Data Grid 7.2

galderz:
https://issues.jboss.org/browse/CLOUD-2683

### DIFF
--- a/official.yaml
+++ b/official.yaml
@@ -308,6 +308,12 @@ data:
       - location: https://raw.githubusercontent.com/jboss-container-images/jboss-datagrid-7-openshift-image/datagrid72/templates/datagrid72-postgresql.json
         docs: https://github.com/jboss-container-images/jboss-datagrid-7-openshift-image/blob/datagrid72-dev/docs/datagrid72-postgresql.adoc
     imagestreams:
+      - location: https://raw.githubusercontent.com/jboss-openshift/application-templates/{xpaas_version}/jboss-image-streams.json
+        docs: https://access.redhat.com/documentation/en/red-hat-jboss-middleware-for-openshift/3/paged/red-hat-jboss-data-grid-for-openshift/
+        regex: jboss-datagrid
+        suffix: rhel7
+        tags:
+          - online-professional
       - location: https://raw.githubusercontent.com/jboss-container-images/jboss-datagrid-7-openshift-image/datagrid72/templates/datagrid72-image-stream.json
         docs: https://access.redhat.com/documentation/en/red-hat-jboss-middleware-for-openshift/3/paged/red-hat-jboss-data-grid-for-openshift/
         regex: jboss-datagrid
--- a/official.yaml
+++ b/official.yaml
@@ -287,8 +287,28 @@ data:
           - online-professional
       - location: https://raw.githubusercontent.com/jboss-openshift/application-templates/{xpaas_version}/datagrid/datagrid71-postgresql.json
         docs: https://github.com/jboss-openshift/application-templates/blob/{xpaas_version}/docs/datagrid/datagrid71-postgresql.adoc
+      - location: https://raw.githubusercontent.com/jboss-container-images/jboss-datagrid-7-openshift-image/datagrid72/templates/datagrid72-basic.json
+        docs: https://github.com/jboss-container-images/jboss-datagrid-7-openshift-image/blob/datagrid72-dev/docs/datagrid72-basic.adoc
+        tags:
+          - online-professional
+      - location: https://raw.githubusercontent.com/jboss-container-images/jboss-datagrid-7-openshift-image/datagrid72/templates/datagrid72-https.json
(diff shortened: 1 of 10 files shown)


official.yaml
```diff
@@ -287,8 +287,28 @@ data:
...
+        tags:
+          - online-professional
+      - location: https://raw.githubusercontent.com/jboss-container-images/jboss-datagrid-7-openshift-image/datagrid72/templates/datagrid72-mysql.json
+        docs: https://github.com/jboss-container-images/jboss-datagrid-7-openshift-image/blob/datagrid72-dev/docs/datagrid72-mysql.adoc
+      - location: https://raw.githubusercontent.com/jboss-container-images/jboss-datagrid-7-openshift-image/datagrid72/templates/datagrid72-postgresql-persistent.json
+        docs: https://github.com/jboss-container-images/jboss-datagrid-7-openshift-image/blob/datagrid72-dev/docs/datagrid72-postgresql-persistent.adoc
+        tags:
+          - online-professional
+      - location: https://raw.githubusercontent.com/jboss-container-images/jboss-datagrid-7-openshift-image/datagrid72/templates/datagrid72-postgresql.json
+        docs: https://github.com/jboss-container-images/jboss-datagrid-7-openshift-image/blob/datagrid72-dev/docs/datagrid72-postgresql.adoc
```
bparees: no tags on this one, intentional?

official.yaml
```diff
@@ -287,8 +287,28 @@ data:
...
     imagestreams:
-      - location: https://raw.githubusercontent.com/jboss-openshift/application-templates/{xpaas_version}/jboss-image-streams.json
+      - location: https://raw.githubusercontent.com/jboss-container-images/jboss-datagrid-7-openshift-image/datagrid72/templates/datagrid72-image-stream.json
```
bparees: this change appears to be removing a bunch of imagestreams. We never remove imagestreams because they may still be referenced by users.  You can mark them hidden but you can't remove them.

luciddreamz:
@bparees What is removing Data Grid 7.1 image stream going to do to existing Data Grid 7.1 deployments on the clusters? Is this also what we're doing in OCP (removing image streams)?

bparees:
> @bparees What is removing Data Grid 7.1 image stream going to do to existing Data Grid 7.1 deployments on the clusters? 

break them.  they won't be able to re-deploy because the imagestream won't resolve.

> Is this also what we're doing in OCP (removing image streams)?

no, we never remove imagestreams, we mark them hidden if we don't want new resources to reference them.  Hence my comment in this PR.

official.yaml
galderz: We're following what others are doing here. E.g.

> docs: https://github.com/jboss-container-images/jboss-datavirt-6-openshift-image/tree/datavirt64-dev/resources/openshift/docs/datavirt64-basic-s2i.adoc

bparees: so it looks like you want to publish the persistent version of the template but not the ephemeral one. that makes sense and is fine.

official.yaml
galderz: I'll fix that

official.yaml
galderz: Not sure what you mean there. 

I'm trying to add new templates for Data Grid 7.2.x, both persistent and ephemeral versions.

galderz:
I've added a new commit that restores previous image stream and just add a new entry instead.

official.yaml
bparees: by publish i mean "provide to online".  You only had the "online-professional" tag on the persistent template, which would be what i'd expect.

official/README.md
```diff
@@ -66,22 +66,10 @@ Docs: [https://github.com/sclorg/dancer-ex/blob/master/README.md](https://github
...
-Path: official/datagrid/imagestreams/jboss-datagrid71-openshift-rhel7.json  
-### jboss-datagrid65-client-openshift
-Source URL: [https://raw.githubusercontent.com/jboss-openshift/application-templates/ose-v1.4.8-1/jboss-image-streams.json](https://raw.githubusercontent.com/jboss-openshift/application-templates/ose-v1.4.8-1/jboss-image-streams.json )  
-Docs: [https://access.redhat.com/documentation/en/red-hat-jboss-middleware-for-openshift/3/paged/red-hat-jboss-data-grid-for-openshift/](https://access.redhat.com/documentation/en/red-hat-jboss-middleware-for-openshift/3/paged/red-hat-jboss-data-grid-for-openshift/)  
-Path: official/datagrid/imagestreams/jboss-datagrid65-client-openshift-rhel7.json  
-### jboss-datagrid71-client-openshift
-Source URL: [https://raw.githubusercontent.com/jboss-openshift/application-templates/ose-v1.4.8-1/jboss-image-streams.json](https://raw.githubusercontent.com/jboss-openshift/application-templates/ose-v1.4.8-1/jboss-image-streams.json )
```
bparees: all these imagestreams are still being deleted by this PR.

galderz: That was automatically done by the `make import` call. No idea why

galderz:
Updated

bparees:
looks good now, thanks

PR #98, opened 2018-07-02
allow for arch specific tags

bparees:


bparees:
@coreydaley ptal.  this will allow us to introduce arch_* tags on content and import it into a "arch/x86" "arch/ppc", etc set of subdirs.

import_content.py
```diff
@@ -263,17 +274,19 @@ def main():
     # parse command line options
     parser = argparse.ArgumentParser(description='Build OpenShift template and image-stream library')
     parser.add_argument("-t", "--tags", nargs='?', help="Select specific tag(s) to import templates/imagestreams (separated by comma ',')")
+    parser.add_argument("-d","--dir",nargs='?',help="Specify a target directory for the imported content")
```
coreydaley: space after commas

import_content.py
```diff
@@ -250,6 +250,17 @@ def has_tag(item, filter_tags):
...
+    if options == None:
+        options=[]
+    # if the tags do not include an architecture indication,
+    # default to x86_64.
+    for tag in options:
+        if tag.startswith('arch_'):
+            arch_found = True
+            break
+    if not arch_found:
+        options.append('arch_x86_64')
```
coreydaley: not sure i feel good about setting this default here.  Maybe we add the arch to the yaml files to get things started?

bparees: It was a combination of not wanting to add the tag to all the existing content and I don't want to police people adding the x86 tag to all new content. It seems like a really reasonable default though I agree it's a bit hacky

bparees:
You can already list multiple arches with this implementation. No need to introduce a new concept and logic.  The only thing that makes it at all special is the x86 defaulting logic, otherwise it's just a tagging convention and the ability to redirect the import to a dir(unrelated to the tag filter you may or may not apply)

PR #100, opened 2018-07-04
Add Fuse entries to Online starter plan

astefanutti:
This PR is meant to make the Fuse entries available in the OpenShift catalog for the Online Starter plan.

Please let me know if there is any additional change to make.

### DIFF
--- a/official.yaml
+++ b/official.yaml
@@ -661,20 +661,24 @@ data:
       - location: https://raw.githubusercontent.com/jboss-fuse/application-templates/GA/quickstarts/spring-boot-camel-template.json
         docs: https://access.redhat.com/documentation/en-us/red_hat_fuse/7.0/html/fuse_on_openshift_guide/
         tags:
+          - online-starter
           - online-professional
       - location: https://raw.githubusercontent.com/jboss-fuse/application-templates/GA/quickstarts/spring-boot-camel-xml-template.json
         docs: https://access.redhat.com/documentation/en-us/red_hat_fuse/7.0/html/fuse_on_openshift_guide/
         tags:
+          - online-starter
           - online-professional
       - location: https://raw.githubusercontent.com/jboss-fuse/application-templates/GA/fis-console-namespace-template.json
         docs: https://access.redhat.com/documentation/en-us/red_hat_fuse/7.0/html/managing_fuse/
         tags:
+          - online-starter
           - online-professional
     imagestreams:
       - location: https://raw.githubusercontent.com/jboss-fuse/application-templates/GA/fis-image-streams.json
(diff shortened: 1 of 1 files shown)


luciddreamz:
Has anyone tried these in OpenShift Online Starter or Pro? I just imported the image streams into OpenShift Online Starter and tried to deploy the console using the provided template and, as I expected after glancing at the template, the resource requests/limits cause the app to fail to deploy. This is probably true in Online Pro as well:

screen shot 2018-07-04 at 10 44 51 am

astefanutti:
@luciddreamz the templates have been tested here: https://issues.jboss.org/browse/FUSEQE-1117.

For the `fuse70-console` template you've tried, it should work with the default memory limit of 32Mi. I'm surprised in your screenshot you have a limit set to 100Mi set. Could it be that it fails because the minimum container/pod size has to be 256MiB as per https://www.openshift.com/products/online/?

luciddreamz:
@astefanutti The lowest memory limit you can specify in OpenShift Online is 100Mi (previously 256Mi, but recently lowered). Setting a memory limit below this value will cause the deployment to fail as shown in the screenshot (the template defaults to 32Mi).

astefanutti:
@luciddreamz ok, let me update the template so that the default memory limit is 100Mi.

luciddreamz:
@astefanutti See updated comment above :)

astefanutti:
@luciddreamz thanks 😉!

astefanutti:
I've just pushed the update on the branch referenced from the catalog: https://github.com/jboss-fuse/application-templates/commit/697aa06c6e5843a71b5fcc4843ad81dc3f8c6339. It should be picked up with the next automatic update. I've tested that the console now works ok on OpenShift Starter Plan.

astefanutti:
@bparees yes, I wasn't sure so I've increased the limit only and tested that it actually works. I would rather then leave the request as the actual amount of memory for the application to run nominally.

I find OpenShift behaviour sub-optimal as even if deployment would specify lower limit, it could just take the (granted limit, app limit) majorant instead of failing. As an end-user, I find it weird that an app that specify a memory limit of 32Mi fails as the minimum limit is 100Mi. I understand this as a way to limit the number of pod running, but in that case just committing 100Mi for the app would do it.

luciddreamz:
@bparees - Increasing the limit should be fine. Everything else will be ignored and set automatically based on that memory limit in OpenShift Online Starter/Pro (tested, confirmed working with only changing the limit).

bparees:
thanks @astefanutti @luciddreamz

astefanutti:
@bparees @luciddreamz thanks for the great help!

PR #101, opened 2018-07-05
[RHDM-662] - Update RHDM templates on OCP and OSO

spolti:


official/rhdm/templates/rhdm70-kieserver-https-s2i.json
```diff
@@ -3,16 +3,16 @@
     "kind": "Template",
     "labels": {
         "template": "rhdm70-kieserver-https-s2i",
-        "xpaas": "1.4.8"
+        "xpaas": "1.4.0"
```
bparees: is this value correct/relevant?

errantepiphany: Honestly I don't even know what this variable is used for... :/

bparees: it's going to apply that label to all the resources created by the template.

i'm going to merge this as is, but if that label isn't relevant, remove it from the upstream template at some point.

spolti: @bparees Thanks for pointing this out, we will remove it for the next releases.

PR #106, opened 2018-08-17
Add EAP 7.2.beta image to official yaml, and correct EAP doc links

luck3y:

### DIFF
--- a/official.yaml
+++ b/official.yaml
@@ -441,21 +441,57 @@ data:
         docs: https://github.com/jboss-openshift/application-templates/blob/{eap_version}/docs/eap/eap71-third-party-db-s2i.adoc
       - location: https://raw.githubusercontent.com/jboss-openshift/application-templates/{eap_version}/eap/eap71-tx-recovery-s2i.json
         docs: https://github.com/jboss-openshift/application-templates/blob/{eap_version}/docs/eap/eap71-tx-recovery-s2i.json
+      - location: https://raw.githubusercontent.com/jboss-container-images/jboss-eap-7-openshift-image/eap72/templates/eap72-mongodb-persistent-s2i.json
+        docs: https://github.com/jboss-container-images/jboss-eap-7-openshift-image/blob/eap72/docs/templates/eap72-mongodb-persistent-s2i.adoc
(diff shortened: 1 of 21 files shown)


luck3y:
Note this picked up some additional, non-EAP product changes, I included those too, as that's typically what we've done in the past.

bparees:
lgtm.  @luciddreamz do we want to start removing the older templates (70, 71, etc?).  

@luck3y also note that you'll need to update your upstream 7.2 imagestreams in the same way i am updating 70/71/etc here:
https://github.com/jboss-openshift/application-templates/pull/475/files#diff-65da8f20834ad10278a40b386213f62aR41

bparees:
> Note this picked up some additional, non-EAP product changes, I included those too, as that's typically what we've done in the past.

yup, that's the process

bparees:
going to merge for now, we can revisit removing the old templates if desired later.

luck3y:
@bparees Yep, imagestream update is OTW for these seperatly, I have PRs open waiting for review.

luck3y:
Oh, and re: 70 and 71, the only thing left of 7.0 is the imagestream, 7.1 is still production, so we shouldn't remove it.

bparees:
> Oh, and re: 70 and 71, the only thing left of 7.0 is the imagestream, 7.1 is still production, so we shouldn't remove it.

ok, I didn't look that closely.  sounds good.

luck3y:
Oh, and final, final not, the EAP 7.2 image will be relaese on Aug 21 next week, so if we could push this right around that time, that would be great. I'll have the other imagestream updates merged and PRs ready here for those too by then.

luck3y:
@bparees When we merge the imagestream change, it should get picked up automatically that night, correct? Will this also do the same for 3.9 / 3.10?

bparees:
> @bparees When we merge the imagestream change, it should get picked up automatically that night, correct?

correct.

> Will this also do the same for 3.9 / 3.10?

this will have no impact on 3.9+3.10, we don't update the imagestreams in previously shipped versions of OCP.

OSO will pick up the change on whatever schedule they do (i'm not sure when they refresh from library).

v3.11 *could* pick it up, but you're going to need to remind us to make sure that happens.

luck3y:
@bparees OK, thanks Ben.

PR #114, opened 2018-09-14
introduce okd, ocp tags

bparees:


bparees:
@coreydaley please review the import_content.py changes

@gabemontero this is a starting point for the tags.  As you work through what is currently in openshift-ansible you'll probably find you need to tweak the "ocp" tags and possibly add entirely new content that's not currently being imported into library today.  In particular we seem to be lacking a lot of the ppc64le content that ansible is distributing.  @jeffdyoung can help you get that content imported into library.

for okd, we should try to make sure everything that's currently in "openshift/origin" has an okd tag on it here.

the commands i ran to import the content were basically:
```
 ./import_content.py --all-tags okd,arch_x86_64 -d operator/okd-x86_64
./import_content.py --all-tags ocp,arch_x86_64 -d operator/ocp-x86_64
./import_content.py --all-tags ocp,arch_ppc64le -d operator/ocp-ppc64le
```

Note that i did not create an okd-ppc64le directory.  There is currently no content tagged for okd+ppc64le.  I don't know if we even want to have any such content.

I'll be updating the jenkins job to keep the "operator" directory up to date on a nightly basis, after this merges.

gabemontero:
Yeah @jeffdyoung and I have already emailed/communicated about him submitting PRs against this repo to inject the ppc64le content.

import_content.py
```diff
@@ -273,14 +280,16 @@ def main():
     """
     # parse command line options
     parser = argparse.ArgumentParser(description='Build OpenShift template and image-stream library')
-    parser.add_argument("-t", "--tags", nargs='?', help="Select specific tag(s) to import templates/imagestreams (separated by comma ',')")
+    parser.add_argument("-t", "--tags", nargs='?', help="Select only content with at least one of the specified tag(s) to import templates/imagestreams (separated by comma ',')")
+    parser.add_argument("--all-tags", nargs='?', help="Select only content with all specified tags to import templates/imagestreams (separated by comma ',')")
```
coreydaley: Can you use --tags and --all-tags together? What should happen if you do?

bparees: you'll get the content that contains at least one of the tags you specified in --tags and contains all the tags you specified in --all-tags.

coreydaley: Maybe --match-all-tags would be better?
so, --tags tag1,tag2 --match-all(-tags) ?

bparees:
broke this out into separate commits for easier review, no actual changes in the content though.

import_content.py
bparees: i think you're bikeshedding but i've pushed the change.

coreydaley:
@bparees lgtm

bparees:
@coreydaley can you put a /lgtm on it? i want to see if i've gotten prow configured properly on this repo so it merges

bparees:
(if @gabemontero has concerns about the tags we can followup)

PR #116, opened 2018-09-14
duplicate core templates under community for operator centos installs

gabemontero:
@bparees ptal ... per a subset of our irc discussion on #aos-devex today this will enable the operator to install the "core" templates along with the centos image streams in the openshift namespace (duplicating what you would see when running `oc cluster up` on v3.11 or earlier)

### DIFF
--- a/community.yaml
+++ b/community.yaml
@@ -53,6 +53,14 @@ data:
         suffix: centos7
         tags:
           - okd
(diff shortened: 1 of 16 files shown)


gabemontero:
I'll be doing the ansible compare in a separate PR

PR #119, opened 2018-09-17
update official.yaml for library/ansible reconciliation; ran import t…

gabemontero:
…o update operator subdir for okd,ocp x86

@openshift/sig-developer-experience ptal

So this attempts to match what 
1) https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_examples/tasks/main.yml does with ...
2) ... the content pulled in from  https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_examples/examples-sync.sh

into openshift/library, specifically the operator subdir, which is what the samples operator in 4.0 will use to install samples in either okd or ocp

@rcernich - if you or someone in xpaas land you delegate to could spend some cycles reviewing the xpaas specifics, that would be great

@sspeiche @luciddreamz heads up to you two as well, given past interest in the sample templates/imagestreams that are installed in various versions of the product

### DIFF
--- a/official.yaml
+++ b/official.yaml
@@ -324,8 +324,12 @@ data:
           - online-professional
       - location: https://raw.githubusercontent.com/jboss-openshift/application-templates/{amq_version}/amq/amq63-persistent-ssl.json
         docs: https://github.com/jboss-openshift/application-templates/blob/{amq_version}/docs/amq/amq63-persistent-ssl.adoc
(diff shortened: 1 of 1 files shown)


bparees:
@gabemontero looks like you need to rebase and rerun make import.

the changes to official.yaml itself look ok to me.

PR #120, opened 2018-09-17
update ips and ds template/imagestreams

spolti:
Signed-off-by: Filippe Spolti <fspolti@redhat.com>


bparees:
i've created a corrected version of your PR here: https://github.com/openshift/library/pull/122

spolti:
I usually do this, but this time I didn't do this because I did not want to include other products in this PR.
Thanks.

PR #124, opened 2018-09-18
Inadvertently used the 1.3 version that is about to be released in sprint24

vhalbert:


bparees:
you seem to have left out the actual official.yaml file change?

PR #125, opened 2018-09-18
Adding back the 6.3 image streams

vhalbert:


official/datavirt/imagestreams/jboss-datavirt64-driver-openshift-rhel7.json
```diff
@@ -41,12 +24,9 @@
                 },
                 "from": {
                     "kind": "DockerImage",
-                    "name": "registry.redhat.io/jboss-datavirt-6/datavirt64-driver-openshift:1.2"
+                    "name": "registry.access.redhat.com/jboss-datavirt-6/datavirt64-driver-openshift:1.2"
```
bparees: you've reverted the registry.redhat.io change.

bparees:
@vhalbert i have a clean version of your PR here:
https://github.com/openshift/library/pull/126

but going back to the "1.2" version reverts the registry.redhat.io change which i don't want to do.  What is the plan for moving this forward while retaining the registry.redhat.io changes?

vhalbert:
Part of the problem is the references used in the official.yaml is to what is currently released, which hasn't been updated with the new registry.redhat.io locations.    So when make import is run, the imagestreams are being updated with the old content.    So not until a new JDV release cant this referencing the correct product changes.

bparees:
ok, well https://github.com/openshift/library/pull/126 got inadvertently merged, so i'm going to close this PR.  When can we expect the new JDV release so we can get the registry hostname fixed?

vhalbert:
Wish there was a way to override the content in library to correct it in the mean time.   But JDv 6.4.4 is in this sprint 24, which we start the release builds tomorrow.    Maybe we can work with Marek Schmidt and get this thru the release quickly, because there's not much to this release.

PR #127, opened 2018-09-25
Update EAP CD streams and templates to point to CD13-OS tag

luck3y:

Note no changes in templates, just adding a tag to fix the source version.


luck3y:
Note, as discussed before, I left any other pulled changes in the PR, these changes are only in official.yaml though.

luck3y:
Thx @bparees!

PR #129, opened 2018-10-04
Update library with latest 1.3 JDV release

vhalbert:


gabemontero:
@bparees @vhalbert this PR only updated the `official` subdir

Would we not want the apply the appropriate tagging and updates so the `operator` subdir gets updated so this content get included in what the 4.0 sample operator uses?

gabemontero:
nevermind @bparees @vhalbert ... forgot the overnight jobs do that

vhalbert:
It was decided that since JDV is no longer being sold, we don't want to expose it unnecessarily to create a new support exposure for someone that is not currently a client.

Van Halbert

----- Original Message -----

PR #130, opened 2018-10-05
3scale: Update templates to use Apicast 2.3

miguelsorianod:


official/3scale/imagestreams/apicast-gateway.json
```diff
@@ -61,6 +61,21 @@
...
+                    "iconClass": "icon-3scale",
+                    "openshift.io/display-name": "3scale APIcast API Gateway 2.3.0.GA",
+                    "openshift.io/provider-display-name": "Red Hat, Inc.",
+                    "tags": "api,gateway,3scale",
+                    "version": "2.3.0.GA"
+                },
+                "from": {
+                    "kind": "DockerImage",
+                    "name": "registry.redhat.io/3scale-amp23/apicast-gateway"
+                },
```
bparees: this needs the local reference policy set on it, see the 2.20 tag above for an example.

miguelsorianod: I've added the local reference policy on version 2.3.0GA

bparees: (you can actually see the 2.3.0 tag is already present: https://github.com/openshift/library/blob/master/official/3scale/imagestreams/apicast-gateway.json#L66-L83)

it also appears to have the reference policy set which i don't quite understand since it appears to be missing from you upstream source.

bparees: well now i see why, these imagestreams are being imported from the master branch of upstream, i don't know if that's intentional or not:
https://github.com/openshift/library/blob/master/official.yaml#L16

bparees: in summary, if it's intentional to have library import from master then this PR is not needed and library already has the content i think you want.

if it's not intentional to have library import from master (e.g. it should be importing from the 2.3.0.GA branch/tag), then you need to update official.yaml to fix the rhamp_version, and you also need to fix the upstream content in that branch so when it gets imported, it is correct (e.g. has the local reference policy set).

miguelsorianod: Hi @bparees .

At this moment is intentional to have the library import from master.
It seems that an automated jenkins build introduced this commit: https://github.com/openshift/library/commit/e73cf4d686c500c08c6209ad6a716272b4c056e2

Which are the same contents than the ones in this PR.

I assume that jenkins job that performed the commit looked at the URL:
      - location: https://raw.githubusercontent.com/3scale/3scale-amp-openshift-templates/{rhamp_version}/3scale-image-streams.yml

defined in the official.yaml file, and because rhamp_version points to master and we updated the master contents of that repository with the changes too then it updated automatically this repository.

This PR was submitted before the jenkins build that commited the changes and that was the reason the changes were not here at the moment of the PR.

I understand then that this PR is not necessary anymore and the issue could be closed if I'm not missing anything.

bparees: yup, sounds right, thanks @miguelsorianod, i'll close this and it sounds like you're all set, at least until you want to come back and replace rhamp_version with a version specific branch/tag instead of master.

PR #141, opened 2018-12-10
Add Fuse Apicurito

astefanutti:
To be merged after #140.

### DIFF
--- a/official.yaml
+++ b/official.yaml
@@ -1028,6 +1028,12 @@ data:
           - ocp
           - online-starter
           - online-professional
+      - location: https://raw.githubusercontent.com/jboss-fuse/application-templates/GA/fuse-apicurito.yml
+        docs: https://access.redhat.com/documentation/en-us/red_hat_fuse/7.2/html/designing_apis_with_apicurito/
+        tags:
+          - ocp
+          - online-starter
+          - online-professional
     imagestreams:
       - location: https://raw.githubusercontent.com/jboss-fuse/application-templates/GA/fis-image-streams.json
         docs: https://access.redhat.com/documentation/en-us/red_hat_fuse/7.2/html/fuse_on_openshift_guide/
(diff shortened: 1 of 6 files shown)


coreydaley:
You need to run "make import" and then commit the results.

astefanutti:
This PR requires PR #140 to be merged first.

astefanutti:
I've rebased the PR as #140 has been merged. CI build fails for discrepancies on 3scale resources which are unrelated to this PR.

bparees:
> CI build fails for discrepancies on 3scale resources which are unrelated to this PR.

if you run make import you won't have discrepancies

@coreydaley i don't think our nightly job is properly importing changes on every run.  i've seen several cases where it reports no changes but if i run make import i see changes.

astefanutti:
Ok I've just added the changes to the 3Scale resources.

bparees:
@astefanutti thanks!

/lgtm

coreydaley:
> @coreydaley i don't think our nightly job is properly importing changes on every run. i've seen several cases where it reports no changes but if i run make import i see changes.

It is possible that something changed after the nightly job.  You would have to run it at about the same time manually to know.  (or is that what you did?)

bparees:
> It is possible that something changed after the nightly job. You would have to run it at about the same time manually to know. (or is that what you did?)

i know because i checked when the upstream commits went in and i looked at the last 20 days of nightly runs which didn't pick it up, then i ran it manually and it picked it up.

but i don't see anything obvious in the nightly job that is wrong, so i'm not sure what to do to investigate it.. it's annoying though because there are cases where we're not picking up upstream changes until someone creates a PR and happens to pick it up.

coreydaley:
Ok, i'll poke around and see if there is any additional logging that can be added.

bparees:
the runs showed git porcelain reporting no delta after running make import.. so either that command is unreliable (unlikely?), the import isn't actually importing in some cases (maybe?) or the git repo is messed up on jenkins (maybe?)

PR #143, opened 2019-01-24
modified the mailcatcher.json template readyness probe

esnoeijs:
Currently the readyness probe is failing for this template because it's only trying to match the `/usr/local/bin/ruby` bit of the command, with the added "-f" flag it will match against the full command that includes the "mailcatcher" command.

### DIFF
--- a/community/mailcatcher/templates/mailcatcher.json
+++ b/community/mailcatcher/templates/mailcatcher.json
@@ -125,6 +125,7 @@
                                     "exec": {
                                         "command": [
                                             "/usr/bin/pgrep",
+                                            "-f",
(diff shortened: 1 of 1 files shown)


bparees:
@esnoeijs that template lives upstream here:

https://github.com/oprudkyi/openshift-templates/blob/master/mailcatcher/mailcatcher.yaml

please get your fix contributed there, this repo just aggregates upstream content, it is not the canonical source for content.

bparees:
(you can see the reference we maintain to the upstream location here: https://github.com/openshift/library/blob/master/community.yaml#L17-L20)

PR #144, opened 2019-01-24
update EAP 7.2 SSO template to use the correct SSO 7.3 adapters

luck3y:


luck3y:
I guess this would probably be auto-imported anyway, but I've opened it just to make sure its not missed.

bparees:
> I guess this would probably be auto-imported anyway, but I've opened it just to make sure its not missed.

in theory yes, though i've seen some issues w/ our auto-importing that i haven't had time to dig into, so it's just as well you submitted this.

luck3y:
@bparees Thanks :)

PR #148, opened 2019-02-15
Add Nginx/Apache (httpd) to Starter/Pro

wgordon17:


gabemontero:
@pweil-  ... looks like @wgordon17 has online users complaining about missing content ... we need someone with root to approve this while @bparees is out (unless he happens to catch this while on PTO) 

Ben is the only approver listed in the OWNERS file

gabemontero:
I did have permissions to initiate https://ci.openshift.redhat.com/jenkins/view/All/job/update_openshift_library/826/

bparees:
@gabemontero since this didn't actually change imported content, the jenkins job is irrelevant in this case.  I think the online team just takes openshift/library and does their own import passing a filter tag (similar to what you do for the samples operator) so i guess they just needed the tag to be there so their import picks up the desired content.

thanks for escalating it and thanks @pweil- 

It's probably reasonable to submit a PR to add @gabemontero @coreydaley and @adambkaplan to the approvers.

PR #150, opened 2019-02-19
add more folks to approver list

bparees:


bparees:
@gabemontero @coreydaley @adambkaplan fyi.  only use your powers for good.

PR #153, opened 2019-02-22
JDG-2121 JDG-2258 Remove deprecated templates, add datagrid service

galderz:
https://issues.jboss.org/browse/JDG-2121
https://issues.jboss.org/browse/JDG-2258

* Deprecated templates have been removed (pre-7.3).
* Caching service template has been removed by the cache service.
* Image stream for 7.2 updated to latest.
* Added datagrid service template.


bparees:
> Why are we keeping the 7.2 S2I image streams? Will there be a 7.3 s2i image version soon?

we generally never remove imagestreams/tags so as to avoid breaking someone who has references to them, during an upgrade.

not removing imagestreamtags is key (since the apply could remove the deleted tags and break refs).

not removing imagstreams is less critical (the upgrade would just ignore the old imagestream) but could still impact people installing imagestreams from here and expecting things that used to exist.  We can delete stuff that's very old/with good reason.

galderz:
Anything else that needs doing here?

PR #161, opened 2019-04-11
chore: add a React Web Example using the Web App builder image

lholmquist:
This adds in a basic react web application example using the Web App Image that we(the nodeshift team) have created.

This also updates any reference to `bucharest-gold` to `nodeshift` now that our team name has changed


adambkaplan:
@siamaksade I'm concerned this naming is too close to `webserver` - thoughts?

lholmquist:
@adambkaplan i think that makes sense,  let me update some things then

lholmquist:
updated with name change

official.yaml
```diff
@@ -75,6 +75,22 @@ data:
...
           - online-professional
+  modern-webapp:
+    templates:
+      - location: https://raw.githubusercontent.com/nodeshift-starters/react-web-app/master/templates/official/basic-react.yaml
+        docs: https://github.com/nodeshift-starters/react-web-app/blob/master/README.md
+        tags:
+          - okd
+          - ocp
+    imagestreams:
+      - location: https://raw.githubusercontent.com/nodeshift/centos7-s2i-web-app/master/imagestreams/webapp-rhel7.json
```
adambkaplan: I think you need to change the imagestream name in nodeshift/centos7-s2i-web-app - looks like the imagestream did not import.

lholmquist: welp,  let me do that also

lholmquist:
pushed another update

lholmquist:
Is there some steps i could take to test this out?  like will it be on a dev cluster somewhere?

adambkaplan:
New image streams should be available in the next beta drop.
We also have nightly and "bleeding edge" payloads: https://openshift-release.svc.ci.openshift.org/

PR #162, opened 2019-04-24
Bump version to 1.4.18 for jboss-amq

kwoodson:
Bumping version to 1.4.18 from 1.4.12.

### DIFF
--- a/official.yaml
+++ b/official.yaml
@@ -1,7 +1,7 @@
 variables:
   xpaas_version: ose-v1.4.8-1
   openjdk_version: ose-v1.4.16
-  amq_version: ose-v1.4.12
+  amq_version: ose-v1.4.18
   eap_version: ose-v1.4.18
   eap_cd_version: CD15
   rhsso72_zstream_version: v7.2.6.GA
--- a/official/README.md
+++ b/official/README.md
@@ -12,29 +12,29 @@ Path: official/3scale/templates/3scale-gateway.json
 # amq
 ## imagestreams
 ### jboss-amq-62
-Source URL: [https://raw.githubusercontent.com/jboss-openshift/application-templates/ose-v1.4.12/amq/amq62-image-stream.json](https://raw.githubusercontent.com/jboss-openshift/application-templates/ose-v1.4.12/amq/amq62-image-stream.json )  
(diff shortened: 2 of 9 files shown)


adambkaplan:
/hold

Is there a critical bugfix in the amq version bump?

adambkaplan:
@kwoodson we inadvertently merged this PR, and subsequently will revert ti in #163.

We are freezing openshift/library until 4.0 GA, with exceptions for critical bugfixes.

kwoodson:
@adambkaplan

This is causing us a small bit of grief. The latest version has been out since October and the official version of the EAP version in the official.yaml is already at  .18. This brings jboss amq in line with EAP.

When does this branch open?

adambkaplan:
Branch will be open latest after GA. I need to verify if we can open after code freeze (May 9th).

bparees:
@adambkaplan depending on what @kwoodson needs (ie if they need these changes in library because they pull from there, but they do not need the changes in the samples operator because they don't use it), it may be time to consider @gabemontero's suggestion of introducing branches in this repo and having the samples operator pull from branches (or tags).

This repo isn't intended to be solely controlled by the samples operator, it's a generic publishing location for content with consumers(samples operator, starter, etc) picking what they want from it.  So we can't hold up publishing for everyone because of the 4.1 release cycle.

gabemontero:
On Fri, Apr 26, 2019 at 8:50 AM Ben Parees <notifications@github.com> wrote:

> @adambkaplan <https://github.com/adambkaplan> depending on what @kwoodson
> <https://github.com/kwoodson> needs (ie if they need these changes in
> library because they pull from there, but they do not need the changes in
> the samples operator because they don't use it), it may be time to consider
> @gabemontero <https://github.com/gabemontero>'s suggestion of introducing
> branches in this repo and having the samples operator pull from branches
> (or tags).
>
Presumably the code at
https://github.com/openshift/cluster-samples-operator/blob/master/tmp/build/library-sync.sh#L7
can be tweaked to reference
a non-master branch/tag.

adambkaplan:
@gabemontero yes it can - tested this out myself on one of the revert branches.

PR #166, opened 2019-04-28
Pointed to new openjdk location and updated xpaas version

sspeiche:
Signed-off-by: Steve Speicher <sspeiche@gmail.com>


adambkaplan:
/lgtm

Note @sspeiche that this likely won't go out in 4.1.0 GA as we've frozen the samples and imagestreams.

PR #170, opened 2019-05-16
Add GoLang CentOS builder image-stream

sspeiche:


community.yaml
```diff
@@ -163,3 +163,10 @@ data:
         suffix: centos7
         tags:
           - okd
+  golang:
+    imagestreams:
+      - location: https://raw.githubusercontent.com/sclorg/golang-container/master/imagestreams/golang-centos7.json
```
adambkaplan: @sspeiche more of an fyi that sclorg is only publishing golang 1.8 images, even though the repo has Dockerfiles for 1.9, 1.10, and 1.11.

PR #176, opened 2019-06-12
[KIECLOUD-214] - Update RHDM and RHPAM 7.3.1 templates on OCP and OSO…

spolti:
… service catalog

Signed-off-by: Filippe Spolti <fspolti@redhat.com>


spolti:
This commit was cherry picker from f7e280f461734f47fa642942cf1fdc87514a11f7
should I run the import_content script and push all changes?

adambkaplan:
@spolti is this intended to update the imagestreams for 4.1.z releases? In general we don't do this.

spolti:
@adambkaplan I cherry picked this commit from master because the rhpam operator version on ocp 4.1 needs the rhpam 7.3.1 image streams.

spolti:
Hi @adambkaplan can this PR be merged?

adambkaplan:
@spolti is this rhpam operator planned to be available on operatorhub.io? (or is it already available?)

spolti:
@adambkaplan it is already there.

adambkaplan:
@spolti an operator distributed on operator hub can't guarantee which imagestreams are deployed on a cluster - in fact an administrator can remove the openshift imagestreams if they so wish.

Is your operator able to fall back to pulling the image via registry.redhat.io?

spolti:
@adambkaplan I guess the operator can do it, @tchughesiv can confirm it.

tchughesiv:
Yes, we fall back to a local image stream pull/creation if it can’t find the image in openshift ns.

tchughesiv:
Of course this relies heavily on the user having created image pull secrets to the registry.

adambkaplan:
/close

Yes, the user will need to provide a pull secret, or import the new imagestream tag themselves to the `openshift` namespace.

sutaakar:
@adambkaplan Considering time schedules and deliveries we cannot deliver image streams with expected product release version for OpenShift to master branch. Considering that RHBA operator is updated on OpenShift when new version is released but image streams aren't, the RHBA operator runs on OpenShift instances which don't contain needed image stream versions.

If I understood correctly this conversation there is no automated way how to make RHBA operator version in sync with image streams, right? There are two options what can user do:
- **manually** import required image streams into openshift namespace
- **manually** create a required pull secret which will be used by image streams generated by RHBA operator to download required images.

Is there any approach which could make sure that RHBA operator and required image streams are available on OpenShift out of the box, without any manual changes?

adambkaplan:
@sutaakar If i'm not mistaken the pull secret used by the installer is placed inside the `openshift-config` namespace, named `pull-secret`. In most cases this will have pull secret to `registry.redhat.io` (note - not guaranteed for disconnected clusters!)

@jwforres can OLM operators use this pull secret to automatically pull images from the Red Hat container catalog?

bparees:
> @jwforres can OLM operators use this pull secret to automatically pull images from the Red Hat container catalog?

Any pod can use that pull-secret because it's supposed to be placed on all the nodes so that pods can cleanly pull from registry.redhat.io w/o user intervention.  Not an OLM-specific thing.  (I am not sure if the openshift-config pull-secret creds are synced to the node, or if there's some other config mechanism that puts the single pull-secret in both openshift-config and  on the nodes, that might be a question for the node/crio team as to where they get the pull-secret from that they put on the nodes).

Only builds are special because they pull images directly, rather than having the kubelet pulling for them, hence they don't get to use the node credentials.

bparees:
to further clarify: 

(1) imagestreams - need a pull-secret in their namespace in order to import and do pullthrough
(2) pods - need a pullsecret in their namespace OR on the node itself in order to pull the image
(3) builds - need a pullsecret in their namespace(and associated w/ the build SA or the buildconfig), or use an imagestream that resides in a namespace that has a pullsecret and uses pullthrough.

if a build or pod is referencing a pullthrough imagestream (rather than directly referencing a pullspec on registry.redhat.io) then really it's just referencing the internal registry, not the external registry, so it falls into (1).

sutaakar:
@adambkaplan @bparees Thanks for info.
To sum it up, if Operator wants to create image streams pointing to Red Hat registry in user namespace, the recommended approach is to copy pull secret named `pull-secret` from `openshift-config` namespace to user namespace?

bparees:
> To sum it up, if Operator wants to create image streams pointing to Red Hat registry in user namespace, the recommended approach is to copy pull secret named pull-secret from openshift-config namespace to user namespace?


While it would work, I would not recommend that as your operator would be exposing the cluster's pull-secret creds to end-users.

If you want to create imagestreams pointing to the redhat registry, in a user namespace, i think you should require the user provide you with creds during/after the operator install.

sutaakar:
ok, thanks

PR #178, opened 2019-06-20
WildFly new image streams and template.

jfdenise:
Starting with WildFly, the image locationhas changed. in addition, we are introducing a new "runtime image" and template.


adambkaplan:
@jfdenise you need to run `make import` in order for the tests to pass.

jfdenise:
@adambkaplan , I did it, then I should commit the modified files and  push?

PR #179, opened 2019-06-29
Update golang CentOS 7 Imagestream Version to 1.10

danielhelfand:
The image used for the golang imagestream is actually `golang 1.10.2`, not `golang 1.8`. The image under the [go-toolset-7-centos7 Dockerhub](https://hub.docker.com/r/centos/go-toolset-7-centos7/tags) is only tagged with `latest` and that image is supporting `1.10.2`.

This can be verified by pulling the image and checking its golang version:

```
$ docker run -it docker.io/centos/go-toolset-7-centos7:latest bash
bash-4.2$ go version
go version go1.10.2 linux/amd64
```

This was originally opened as a pr to [sclorg/golang-container](https://github.com/sclorg/golang-container/pull/28), but I'm not sure who if anyone is maintaining that repo anymore. So I thought it made sense to try opening this here.

### DIFF
--- a/community/golang/imagestreams/golang-centos7.json
+++ b/community/golang/imagestreams/golang-centos7.json
@@ -10,8 +10,9 @@
     "spec": {
         "tags": [
             {
+                "name": "latest",
                 "annotations": {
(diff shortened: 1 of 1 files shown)


adambkaplan:
/close

openshift/library automatically syncs its content from upstream imagestream repos. Existing PR there should be sufficient.

danielhelfand:
@adambkaplan Thanks! Wasn't sure whose attention to get over in `sclorg`.

PR #186, opened 2019-08-15
Disable travis ci

coreydaley:
- Remove .travis.yml file
- Update make verify to run make dep to install dependencies

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,0 @@
-language: python
-python:
-  - "2.7"
-  - "3.6"
-install:
-  - "pip install -U pip"
-  - "pip install -r requirements.txt"
-script: make verify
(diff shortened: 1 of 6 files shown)


coreydaley:
/assign @bparees 

Ben, Can you also disable Travis CI for this repository?

Makefile
```diff
@@ -1,6 +1,6 @@
 .DEFAULT_GOAL := help
 
-verify: ## Run core verification. Example: make verify
+verify:  dep ##Run core verification. Example: make verify
```
bparees: ?

coreydaley: That runs make dep when you run make verify

bparees:
> Ben, Can you also disable Travis CI for this repository?
done

Makefile
bparees: yeah....why would we do that?  i wouldn't expect a "Verify" to change my content.

(also put it on its own line, like we have elsewhere, separate from the comment)

coreydaley: We need to install the dependencies for the python script that does the import?

bparees: i don't need to do that if i have them installed.

that sounds like a pre-req step, not a verify step.

Makefile
```diff
@@ -1,6 +1,6 @@
 .DEFAULT_GOAL := help
 
-verify: ## Run core verification. Example: make verify
+verify: ##Run core verification. Example: make verify
```
bparees: drop this change

coreydaley: updated in current push

PR #188, opened 2019-09-09
update verify image so it can perform samples importing+pushing

bparees:


bparees:
/assign @adambkaplan 

needed for https://github.com/openshift/release/pull/4971  (well, not strictly needed, I already built+pushed the image somewhere, but to keep things nicely in sync...)

bparees:
@adambkaplan fyi i pushed this image to quay.io/bparees/importer and that's how i'm using it in the new job.  If i can make the job use the image directly from api.ci i will, but since i'm creating a prow job (not a ci-operator job) i think the use of imagestreams is a bit more constrained.

PR #189, opened 2019-09-13
Add Open Liberty imagestream to official imagestreams

brenthdaniel:
Add an imagestream for open liberty s2i to the list of official image streams.


adambkaplan:
@brenthdaniel thanks for your PR! We're happy to have OpenLiberty join the family of imagestreams available on OpenShift.

Per our [guidelines](https://github.com/openshift/library#official), `official` is only reserved for images provided and supported by Red Hat. Please update the PR to file the imagestream under the `community` group and we'll be glad to add it in.

adambkaplan:
/hold

For move to `community`

brenthdaniel:
Hi, 
  I chose "official" based on the fact that RH will be supporting  Open Liberty (at least on a first level basis). I'm not sure of the implications of "official" vs "community". If we deliver to community now, would we be able to move it when the support statement gets finalized?

bparees:
> If we deliver to community now, would we be able to move it when the support statement gets finalized?

yes, moving it is no problem.

the distinction for us is "if a customer contacts us with a problem with this image/imagestream, is there a team at redhat dedicated to supporting that customer and providing a fix".

brenthdaniel:
Thanks, I updated the PR to move it to community.

brenthdaniel:
@bparees Is there anything else that needs to be done here prior to merging?

PR #196, opened 2019-12-04
Update ImageStreamTag name to version only

jmtd:
This is a copy of https://github.com/jboss-container-images/openjdk/pull/103 which was a problem for OCP 4.3 (see https://bugzilla.redhat.com/show_bug.cgi?id=1778613)

I've targetted `master` but I'm not sure which versions of OCP are affected nor which branches are relevant to those. Can you recommend whether we need to merge to other places too?

### DIFF
--- a/official/java/imagestreams/java-rhel7.json
+++ b/official/java/imagestreams/java-rhel7.json
@@ -67,7 +67,7 @@
                 },
                 "from": {
                     "kind": "ImageStreamTag",
-                    "name": "java:11"
+                    "name": "11"
                 },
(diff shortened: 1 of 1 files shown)


gabemontero:
/close 

see my comment https://github.com/jboss-container-images/openjdk/pull/103#issuecomment-561689644 @jmtd 

we don't pull in these changes via the manner you've done here

cut a new tag with the commit just merged, and we'll update https://github.com/openshift/library/blob/master/official.yaml#L3

PR #190, opened 2019-09-13
remove lone ppc piece of content

gabemontero:
per recent discussions in slack wrt 4.3 push 2 prod, non-x86, epic work items 

nuke the one lone ppc layover from 3.10

@bparees @openshift/openshift-team-developer-experience PTAL

PR #201, opened 2020-01-16
add s390 and ppc imagestream defs for mariadb

gabemontero:
See https://projects.engineering.redhat.com/browse/RHELPLAN-28399

@openshift/openshift-team-developer-experience @bparees @siamaksade fyi

### DIFF
--- a/official.yaml
+++ b/official.yaml
@@ -254,6 +254,20 @@ data:
           - ocp
           - online-starter
           - online-professional
+      - location: https://raw.githubusercontent.com/sclorg/mariadb-container/master/imagestreams/mariadb-rhel7-s390x.json
+        docs: https://github.com/sclorg/mariadb-container/blob/master/README.md
+        regex: mariadb
+        suffix: rhel7-s390x
+        tags:
+          - ocp
+          - arch_s390x
+      - location: https://raw.githubusercontent.com/sclorg/mariadb-container/master/imagestreams/mariadb-rhel7-ppc64le.json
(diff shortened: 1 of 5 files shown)


official/mariadb/imagestreams/mariadb-rhel7.json
```diff
@@ -15,7 +15,7 @@
                     "iconClass": "icon-mariadb",
                     "openshift.io/display-name": "MariaDB (Latest)",
                     "openshift.io/provider-display-name": "Red Hat, Inc.",
-                    "tags": "database,mariadb"
+                    "tags": "database,mariadb,ppc64le"
```
bparees: this looks odd to me?

bparees: (likely a bug in the import code?)

gabemontero: it does .... I think I see the problem (and realize I was not entirely aware of what `make import` did previously)

gabemontero: interestingly enough, the same think existed with the hold ruby ppc entry that existed back in 3.10 .. see https://github.com/openshift/library/pull/190/files

gabemontero: nope it was a problem in the new offical.yaml change ... fwiw the old 3.10 PR was wrong as well

PR #205, opened 2020-01-22
Add all ppc64le and s390x imagestreams

yselkowitz:
This PR includes the results of the import script as would occur on the next automatic sync.


yselkowitz:
/assign @gabemontero 
This blocks your preferred solution to https://github.com/openshift/cluster-samples-operator/pull/222

bparees:
@yselkowitz   we can trigger the nightly import job after this merges if timing is critical, you don't need to include that imported content in your PR.  just running "make import" and committing those results is preferred.  Otherwise you are updating content that might not align w/ what our nightly job does and your changes will get stomped by the nightly job, leading to confusion.

I also see that in some cases we have distinct imagestreams for each architecture(good!) and in other cases we do not(not so good).  In order to have implied buy-in from the SCL team i would expect one of the following:

1) the imagestreams are explicitly annotated w/ their preferred architecture (`importer.image.openshift.io/prefer-arch` annotation on the tag in question)
2) explicit distinct imagestream yaml exists for each architecture

yselkowitz:
Including the import results for x86_64 too (EAP 7.2 changed since the last automatic import) in order to allow the tests to pass.

gabemontero:
and yep the verify job is complaining that the `make import` and `make verify` flow is not right.

So to slightly elaborate on @bparees comment:
- you update only `official.yaml`, pointing to existing multiarch files upstream (i.e. the SCL repos)
-- if the upstream repos do not container those files, PRs against those repos must be filed to add them
- run `make import`
- run `make verify`
- submit PR
- overnight job will populate the `operator` subdir
- after that jobs runs, use of library-sync in a sample operator PR can pull the content in

yselkowitz:
@bparees originally (iow for OCP 3.10) we added separate imagestreams files for ppc64le because the main ones all still included pre-multi-arch images which were not available for ppc64le.  However, many of those pre-multi-arch images are now EOL and have since been removed from the main imagestreams files.  Furthermore, the maintenance burden for three files is obviously much more than one (OpenJDK has already complained about that, although at the moment they still have old tags, so there was no choice in the meantime.)  What technical need is there for a separate imagestreams file when the main one only has multi-arch image tags?

yselkowitz:
Reworked the commits to only include the results of `make import` and `make verify`, but for the record I was using the exact same steps as the job for the other directories.  Once this merges, please do trigger an import job as offered above, as it is a prerequisite of https://github.com/openshift/cluster-samples-operator/pull/222.

yselkowitz:
The tests now pass.

official.yaml
```diff
@@ -197,6 +265,9 @@ data:
           - ocp
           - online-starter
           - online-professional
+          - arch_ppc64le
```
gabemontero: The jenkins images are unique, and as you can see, from openshift itself.  Unless multi arch QE has done any testing on ppc and s390, I can assure everyone that openshift development (the devtools team own now, after I owned them for a few years) or QE has not done testing on those platforms.

@yselkowitz - do you have explicit confirmation that the jenkins related images have been tried on s390/ppc64le ?

Also note, on 4.x they technically overriden by the install payload anyway.  But again, I've never heard of anyone trying it on those platforms.

I'd be inclined to not include these, at least initially, until some due diligence can be done, if it has not been done already.

@akram FYI

official.yaml
```diff
@@ -205,6 +276,9 @@ data:
...
```
gabemontero: same here

official.yaml
```diff
@@ -213,6 +287,9 @@ data:
...
```
gabemontero: same here

official.yaml
```diff
@@ -155,39 +205,57 @@ data:
...
+          - arch_s390x
+          - arch_x86_64
       - location: https://raw.githubusercontent.com/openshift/jenkins/master/openshift/templates/jenkins-persistent-monitored.json
         docs: https://github.com/openshift/origin/blob/master/examples/jenkins/README.md
         tags:
           - okd
           - ocp
           - online-starter
           - online-professional
+          - arch_ppc64le
```
gabemontero: see below

official.yaml
```diff
@@ -155,39 +205,57 @@ data:
...
+          - arch_s390x
+          - arch_x86_64
       - location: https://raw.githubusercontent.com/openshift/jenkins/master/openshift/templates/jenkins-ephemeral-monitored.json
         docs: https://github.com/openshift/origin/blob/master/examples/jenkins/README.md
         tags:
           - okd
           - ocp
+          - arch_ppc64le
```
gabemontero: see below

official.yaml
```diff
@@ -155,39 +205,57 @@ data:
...
+          - arch_s390x
+          - arch_x86_64
   jenkins:
     templates:
       - location: https://raw.githubusercontent.com/openshift/jenkins/master/openshift/templates/jenkins-ephemeral.json
         docs: https://github.com/openshift/origin/blob/master/examples/jenkins/README.md
         tags:
           - okd
           - ocp
+          - arch_ppc64le
```
gabemontero: see below

yselkowitz:
> There have been discussion in our planning session about the need for annotation / separate file (which as noted, was a convention established during 3.10).

The "convention" was done for a specific reason, as noted above.  When that reason no longer applies, as is increasingly the case, we would need to know what other technical reason would justify the continued need to maintain three files instead of one.

> There is still the question of actually hearing from Honza or one of the folks on their team that they are ready to address any issues discovered running their images on s390 / ppc64le openshift.

Those SCL images which are multi-arch are documented as supported as such, and again, there really is nothing new here:

https://access.redhat.com/support/policy/updates/rhscl-rhel7

gabemontero:
> > There have been discussion in our planning session about the need for annotation / separate file (which as noted, was a convention established during 3.10).
> 
> The "convention" was done for a specific reason, as noted above. When that reason no longer applies, as is increasingly the case, we would need to know what other technical reason would justify the continued need to maintain three files instead of one.
> 
> > There is still the question of actually hearing from Honza or one of the folks on their team that they are ready to address any issues discovered running their images on s390 / ppc64le openshift.
> 
> Those SCL images which are multi-arch are documented as supported as such, and again, there really is nothing new here:
> 
> https://access.redhat.com/support/policy/updates/rhscl-rhel7

As recently as with the mysql v8 image, there have been issues with running these in openshift vs. someone setting them up manually on a node.

The direction I'm getting from my end is we want explicit declaration from them about agreeing to support running these images on 4.x s390/ppc64l3 openshift.

yselkowitz:
ping @hhorak

official.yaml
yselkowitz: > The jenkins images are unique, and as you can see, from openshift itself. Unless multi arch QE has done any testing on ppc and s390, I can assure everyone that openshift development (the devtools team own now, after I owned them for a few years) or QE has not done testing on those platforms.

Nothing new there, OpenShift as a whole is in the same situation, and Jenkins is being delivered therewith on Power and Z.

> @yselkowitz - do you have explicit confirmation that the jenkins related images have been tried on s390/ppc64le ?

We understand that testing this is our responsibility, as with everything else.

official.yaml
```diff
@@ -120,6 +158,9 @@ data:
           - ocp
           - online-starter
           - online-professional
+          - arch_ppc64le
+          - arch_s390x
```
hhorak: In https://github.com/sclorg/mariadb-container/pull/115 @gabemontero explained us that we need a separate imagestream files for alt-arch image streams. Of course this way it would be much easier to support, I just don't get where we need a separate file and where we just add tags.

hhorak:
@yselkowitz the changes look correct to me, with just the comment inline, that it is not clear to me whether it was intentional to use separate image stream files somewhere and only more tags elsewhere.

gabemontero:
remove the jenkins updates for now @yselkowitz as previously mentioned `official.yaml`, re-run `make import` and `make verify`, and I'll lgtm this.

based on the discussions leading up to, and the call on Friday, hopefully we'll get to a resolution on what to do with the cluster samples PRs, an possibly adding in jenkins, based on the results of either the devtools team or your team trying out the payload image

gabemontero:
And the clarify (@bparees and I have confirmed this)
- the separate files is not required for homogeneous installs (meaning the cluster is all one architecture type
- the understanding is that if/when we support heterogenous clusters, we will need separate files and the annotation

yselkowitz:
We haven't even started thinking about heterogeneous clusters yet, so it's a bit premature to plan for that.  That being said, to what annotation do you refer?

gabemontero:
> We haven't even started thinking about heterogeneous clusters yet, so it's a bit premature to plan for that. 

Yes, that is why we are willing to merge without the separate file stuff at this time.  Just remove the update you made for jenkins.

That being said, to what annotation do you refer?

Examples:
https://github.com/sclorg/mariadb-container/blob/master/imagestreams/mariadb-rhel7-s390x.json#L19
https://github.com/sclorg/mariadb-container/blob/master/imagestreams/mariadb-rhel7-ppc64le.json#L19

bparees:
>  That being said, to what annotation do you refer?

`importer.image.openshift.io/prefer-arch` which can be put on the imagestream.spec.tag metadata.

@gabemontero @dmage that annotation (and `importer.image.openshift.io/prefer-os`) do not appear to be documented, we should add docs in our imagestream documentation about this.

gabemontero:
sorry ... posted wrong annotation .... see @bparees 's comment https://github.com/openshift/library/pull/205#issuecomment-577765347

PR #207, opened 2020-02-04
[NODE-518] Move modern-webapp from official to community

yselkowitz:
Filing on behalf of @lholmquist.


gabemontero:
@yselkowitz is `[NODE-518]` by chance reference to a Jira somewhere ?  And perhaps that explains the reasoning for the change ?

thanks

lholmquist:
The web-app image that was being used, was only a tech preview and is now being deprecated.  So that is why it needs to be removed from the official listing.  It is being replaced with a community only image

gabemontero:
thanks for the explanation

/lgtm

gabemontero:
ah ... the new 4.4 post feature freeze need a bugzilla requirement is hitting openshift/library as well.

I'll see if that is easy to turn off / something we would want to do and report back.

gabemontero:
@adambkaplan is trying to remove the bz requirement here via https://github.com/openshift/release/pull/7035

PR #208, opened 2020-03-18
[KIECLOUD-369] - Update RHPAM and RHDM templates and imagestreams on …

spolti:
…OpenShift Service Catalog for release 7.7.0


gabemontero:
fyi @openshift/openshift-team-developer-experience another instance of middleware related imagestreams, named along versions, that will no longer be installed in new releases (which lead to the need for https://bugzilla.redhat.com/show_bug.cgi?id=1811143 's fix)

PR #87, opened 2018-05-03
Updated JWS image streams and templates

bekionel:
@sspeiche @luciddreamz Can you please take a look?

PR #225, opened 2020-06-25
Update webserver to 5.3 and add s390x

yselkowitz:
CC: @bekionel please review


official.yaml
```diff
@@ -10,7 +10,8 @@ variables:
   rhsso74_zstream_version: v7.4.0.GA-RHBZ-1813894-fix
   webserver_version: ose-v1.4.17
   webserver3_version: jws31-v1.4
-  webserver5_version: jws50-v1.2
```
bekionel, line 13: Thank you @yselkowitz for the PR. Referencing the discussion in https://github.com/openshift/library/pull/87 , I would say that we shouldn't remove the imagestreams that were included with jws50. We could create a new variable, e.g.   webserver53_version: jws53-v1.0 to reference the newer imagestreams and templates, and also keep the former ones. What do you think? @bparees could you also please advise if I'm wrong? Thanks!

gabemontero:
apologies, but a one time switch over from import being python based to golang based in openshift/library's internal plumbing means you'll need to rebase, and then run `make build` and then `make import` again.

@coreydaley keep me honest here if I have the reconciliation process incorrect

official.yaml
yselkowitz: 5.0 (and 5.2) are no longer supported, only 5.3 is.  We shouldn't be encouraging or enabling use of unsupported versions.

bekionel: I misformulated my reply, I should have mentioned that we should mark the unsupported (5.0) version as hidden. The 5.2 version was never imported so we don't need to do something atm, but since 5.0 was included, it cannot be removed. At least that's what I understood from the conversation in https://github.com/openshift/library/pull/87. If the policy has changed since then, I'm completely OK with your PR as it is. I agree with you that we shouldn't be encouraging use of unsupported versions. @gabemontero perhaps you have more information on this?

gabemontero: yeah things have evolved a bit since 2018 @bekionel 

in 4.x, during an upgrade, samples operator will not remove older imagestreams installed from previous releases which are not part of the latest release, to stay "philosophically aligned" with the discussion you noted in https://github.com/openshift/library/pull/87

it is then up to the cluster administrator to remove unsupported content for his users as he or she sees fit as after deployments / builds using those older versions are migrated 

As such you should feel free to no longer included unsupported content here.  And then new installs (vs. upgrades) will not have the unsupported content.

yselkowitz: @bekionel would you like me to remove 3.0 for you too, thereby leaving only 3.1 and 5.3?

bekionel: @yselkowitz that would be great, thanks! You will spare me one extra PR :)

coreydaley:
@gabemontero I do believe that should work.

coreydaley:
@yselkowitz If you wouldn't mind, would you add a commit to your pull request that removes the following line from the README.md?

`[Build Status](https://ci.openshift.redhat.com/jenkins/job/update_openshift_library/)`

That build system is deprecated and I keep forgetting to remove it.

yselkowitz:
Anything else preventing us from getting /approved /lgtm ?

Issue #227, opened 2020-06-26
Missing content from operator directories


yselkowitz:
An erroneous fix for #226 has removed almost everything from the operator directories:

https://github.com/openshift/library/tree/master/operator/ocp-x86_64/official

/assign @coreydaley

gabemontero:
When I look at https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/log?job=periodic-openshift-library-import&id=1276546179903000576

that came from the run @coreydaley that started at 12:01:41

which I think was about 20 minutes before your openshift/release PR merged 

and yep that deleted a bunch of stuff

to your question https://github.com/openshift/release/pull/9943#issuecomment-650234957 any samples operator PR will run image-eco, jenkins, and build e2e's .... which touch upon a number of samples

yselkowitz:
https://github.com/openshift/release/pull/9943 was merged at 11:02:39 at which time @coreydaley said that the import would run within the hour, which it did.

gabemontero:
thanks for the confirm @yselkowitz 

@coreydaley is working on a fix now .... I'm having him run the periodic job command in his fix openshift/library fix branch and post the results as a separate commit for us to review

and if the batch/periodic commands have to change to get the content back, post that as well in the PR for reference, so we can cross ref with any openshift/release change he has to make

gabemontero:
ok the periodic just ran ( https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/log?job=periodic-openshift-library-import&id=1276607614913875968 ) and no deletes 

along with the (re)additions from https://github.com/openshift/library/pull/228 I think we are good to go but will leave this open and wait for @yselkowitz confirmation

yselkowitz:
Comparing the current contents of master to 98199662db (just before the python->golang transition), the only remaining deletions in operator/ocp-$ARCH seem to be the index.json and README.md files.  The imagestreams and templates themselves are radically reformatted, so it's difficult to verify the differences, but at least the content appears to be present again.

PR #228, opened 2020-06-26
Fix import tag comparison

coreydaley:


gabemontero:
ok it looks like the removed content is added back in

we are going to merge in the readded content with this PR directly

@yselkowitz if you have the cycles by all means cross check us

@coreydaley followed the precise recipe from the batch/periodic, with a separate commit when each command produced a change

once this merges we'll watch the periodic at https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/?job=periodic-openshift-library-import and confirm there are not more massive deletes

gabemontero:
We'll leave https://github.com/openshift/library/issues/227 open until minimally the periodic runs in about 45 minutes and for @yselkowitz to chime in when he gets the chance in case we missed something

PR #230, opened 2020-07-02
Add support for configuration files and cache url fetches

coreydaley:
This pull request:
 - Allows multiple imports to run concurrently by utilizing a configuration file
 - Pre-caches all urls that will be used so that they are only downloaded once.

The entire library periodic job for the cluster-samples-operator now runs in less than three seconds, down from 10-12 seconds.
It also only hits each url once to download instead of once for each import (or once per document per import if there is overlapping urls)


gabemontero:
Are you planning to add `make verify-periodic` to the verify test at https://github.com/openshift/release/blob/master/ci-operator/config/openshift/library/openshift-library-master.yaml#L30 @coreydaley ?

In the interim, did you run that manually and confirm things are OK?

Otherwise, I had no comments on the code changes.

coreydaley:
> Are you planning to add `make verify-periodic` to the verify test at https://github.com/openshift/release/blob/master/ci-operator/config/openshift/library/openshift-library-master.yaml#L30 @coreydaley ?
Yes, I was planning on updating the periodic job in openshift/release to use the new functionality and run the verify-periodic script 
> In the interim, did you run that manually and confirm things are OK?
Yes, i have been running it manually to confirm, and running it the "old way" just to make sure there are no changes to the data.
> Otherwise, I had no comments on the code changes.
Awesomesauce

PR #250, opened 2020-09-10
Remove dotnet templates

yselkowitz:
This is on behest of the .NET team, which has deprecated these and are
not updating them for currently supported versions.

/cc @tmds
/assign @gabemontero


tmds:
@yselkowitz thanks!

PR #252, opened 2020-09-14
[KIECLOUD-418] - Update RHPAM and RHDM templates and imagestreams on …

spolti:
…OpenShift Service Catalog for release 7.8.1

### DIFF
--- a/official.yaml
+++ b/official.yaml
@@ -13,8 +13,8 @@ variables:
   webserver5_version: jws53-v1.0
   webserver5_openj9_version: jws53openj9el7-v1.1
   ips_ds_version: 6.4.12.GA
-  rhdm_version: 7.8.0.GA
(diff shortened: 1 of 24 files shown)


gabemontero:
@spolti am I correct in seeing that all the PHPAM/PHDM imagestream images are rhel/ubi v8 based?

For the products which still have the mixture, we have started asking for adherence to new set of naming conventions, but I think if you all are v8 only we won't need to bother.  But if I'm missing some v7 based images somewhere let me know.

@yselkowitz FYI

yselkowitz:
They made separate imagestreams for the rhel8 based versions some time ago so I'd say it doesn't apply to them.

spolti:
@gabemontero yes, all images are based on ubi8.

PR #255, opened 2020-09-18
feat:  Update the mongodb ephemeral application to use postgresql.

lholmquist:
This also updates the example to use more modern javascript features.

This is part of [NODE-601](https://issues.redhat.com/browse/NODE-601) as well as other internal discussions about replacing this example with an example that is supported by the Nodeshift team

cc @yselkowitz

### DIFF
--- a/official.yaml
+++ b/official.yaml
@@ -93,7 +93,7 @@ data:
     templates:
       - location: https://raw.githubusercontent.com/sclorg/nodejs-ex/master/openshift/templates/nodejs.json
         docs: https://github.com/sclorg/nodejs-ex/blob/master/README.md
-      - location: https://raw.githubusercontent.com/nodeshift-starters/nodejs-rest-http-crud/master/.openshift/template.json
+      - location: https://raw.githubusercontent.com/nodeshift-starters/nodejs-rest-http-crud/master/.openshift/postgresql-template.json
         docs: https://github.com/nodeshift-starters/nodejs-rest-http-crud/blob/master/README.md
         tags:
           - okd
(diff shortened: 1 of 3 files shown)


yselkowitz:
@lholmquist thanks for this!  FYI there are generally two templates provided for database-consuming examples -- including nodejs-mongodb -- one for ephemeral storage (which uses emptyDir) and one for persistent storage (which uses a PVC and has a VOLUME_CAPACITY variable).  The differences between the two are minor, as can be seen from a diff of either your nodejs-mongodb examples or another SCLs.  Now that you have one, could you make the other so that both can be replaced?

gabemontero:
/hold

@yselkowitz - were you and @lholmquist hoping to get this change into 4.6 ?

as this change removes a template that is tested in image-ecosystem, there will be some https://github.com/openshift/origin staging that will be required.

1) before we pull this into cluster-samples-operator, the test for nodejs-mongodb-example will have to be commented out at https://github.com/openshift/origin/blob/master/test/extended/image_ecosystem/sample_repos.go#L150-L162

2) then we merge a change in cluster-samples-operator that will remove `nodejs-mongodb-example` and add `nodejs-postgresql-example`

3) once that cluster-samples-operator PR merges, we should go back to openshift/origin and change https://github.com/openshift/origin/blob/master/test/extended/image_ecosystem/sample_repos.go#L150-L162 to reference the new `nodejs-postgresql-example` template

To do all that at this time, we'll need a BZ (is the ubi8 one still open).

If this is waiting until 4.7, we still have to do those steps above, but will not need a BZ once master opens up for 4.7

yselkowitz:
The goal was 4.6 in tandem with the deprecation of the mongodb imagestream.  I can pick this up next week when I'm back.

lholmquist:
@yselkowitz i know we are waiting on something else,  but i just added the persistent template to this PR also, for when we are ready

coreydaley:
@lholmquist You need to run "make import" locally and then commit the changes.

yselkowitz:
Filed: https://github.com/openshift/origin/pull/25528

yselkowitz:
@lholmquist make import needs to be run again now that https://github.com/nodeshift-starters/nodejs-rest-http-crud/pull/191 has merged.

PR #263, opened 2020-11-23
Updating imagestreams and templates for JWS 5.4

bekionel:
This updates the RHEL7 imagestreams and templates for JWS 5.4 and also adds the UBI8 imagestreams and templates.

### DIFF
--- a/arch/aarch64/official/ruby/imagestreams/ruby-rhel-aarch64.json
+++ b/arch/aarch64/official/ruby/imagestreams/ruby-rhel-aarch64.json
@@ -26,7 +26,29 @@
 				},
 				"from": {
 					"kind": "ImageStreamTag",
-					"name": "2.6-ubi8"
+					"name": "2.7-ubi8"
+				},
+				"generation": null,
+				"importPolicy": {},
+				"referencePolicy": {
+					"type": "Local"
+				}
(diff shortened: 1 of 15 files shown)


bekionel:
For some reason the make import command created the following template in the official/webserver/templates/ dir: jws**53**-openjdk11-tomcat9-rhel7-https-s2i.json whereas it should be creating jws**54**-openjdk11-tomcat9-rhel7-https-s2i.json , and the make verify was successful. I considered that a random glitch so I renamed the template, committed and pushed the changes (without running the make verify, thinking that it wouldn't be needed). The variable is set here: https://github.com/openshift/library/pull/263/commits/63c873d04d4f787b256cd10369b76d6af2581107#diff-9ec8f74fc97e0286c356fe95abf0e2003e3ba8ef2b03841b51d8009e700e5bbbL13 and the actual name of the template comes from here: https://github.com/openshift/library/pull/263/commits/63c873d04d4f787b256cd10369b76d6af2581107#diff-9ec8f74fc97e0286c356fe95abf0e2003e3ba8ef2b03841b51d8009e700e5bbbL1296 so I am a bit lost tbh to what is causing the 53 to appear. Any thoughts/comments welcome!

PR #271, opened 2021-03-08
community: replace some docker.io images with quay.io

vrutkovs:
This would ensure OKD clusters are hitting docker.io less often. Centos8 images are not yet mirrored to quay

### DIFF
--- a/arch/x86_64/community/mariadb/imagestreams/mariadb-centos.json
+++ b/arch/x86_64/community/mariadb/imagestreams/mariadb-centos.json
@@ -64,7 +64,7 @@
 				},
 				"from": {
 					"kind": "DockerImage",
-					"name": "docker.io/centos/mariadb-103-centos7:latest"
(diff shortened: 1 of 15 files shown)


vrutkovs:
These changes need to happen in respective repos

arch/x86_64/community/mariadb/imagestreams/mariadb-centos.json
```diff
@@ -64,7 +64,7 @@
 				},
 				"from": {
 					"kind": "DockerImage",
-					"name": "docker.io/centos/mariadb-103-centos7:latest"
+					"name": "quay.io/centos7/mariadb-103-centos7:latest"
```
gabemontero, line 67: Yeah to echo my comment in the sample operator PR, you cannot just change the value here.

You need to got to https://github.com/sclorg/mariadb-container and change the value there.

And this then applies to the other changes you've made here.

/hold

gabemontero:
> These changes need to happen in respective repos

yeah there you go :-)

PR #274, opened 2021-04-21
Remove golang imagestreams

yselkowitz:

PR #277, opened 2021-05-20
Revert "Remove golang imagestreams"

yselkowitz:


gabemontero:
So revisiting https://github.com/openshift/library/pull/274#issuecomment-824280461 and the catalog pages referenced there, I no longer see mention about using rpms instead.

I take it an about face has occurred with respect to the images going away entirely.

FWIW the removal via #274 missed the sample bump we did for 4.8.  So the golang imagestreams are still there.

/lgtm

PR #280, opened 2021-06-25
Remove tags from mongodb imagestream and nodejs-mongodb template

yselkowitz:
The mongodb imagestream and all dependent thereon was deprecated or
removed in 4.6, with only the minimum required to support the testsuites
remaining.  With the tests no longer dependent on mongodb, now the rest
can be removed.

### DIFF
--- a/official.yaml
+++ b/official.yaml
@@ -114,12 +114,6 @@ data:
         docs: https://github.com/sclorg/nodejs-ex/blob/master/README.md
       - location: https://raw.githubusercontent.com/sclorg/nodejs-ex/master/openshift/templates/nodejs-mongodb.json
         docs: https://github.com/sclorg/nodejs-ex/blob/master/README.md
-        tags:
-          - okd
-          - ocp
-          - arch_ppc64le
-          - arch_s390x
-          - arch_x86_64
       - location: https://raw.githubusercontent.com/nodeshift-starters/nodejs-rest-http-crud/master/.openshift/postgresql-template.json
         docs: https://github.com/nodeshift-starters/nodejs-rest-http-crud/blob/master/README.md
         tags:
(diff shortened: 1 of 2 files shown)


yselkowitz:
Waiting on pretesting via https://github.com/openshift/cluster-samples-operator/pull/379
/hold

yselkowitz:
Pretesting deemed successful: https://github.com/openshift/cluster-samples-operator/pull/379#issuecomment-870857546
/hold cancel

official.yaml
```diff
@@ -114,12 +114,6 @@ data:
         docs: https://github.com/sclorg/nodejs-ex/blob/master/README.md
       - location: https://raw.githubusercontent.com/sclorg/nodejs-ex/master/openshift/templates/nodejs-mongodb.json
         docs: https://github.com/sclorg/nodejs-ex/blob/master/README.md
-        tags:
```
gabemontero, line 117: I'm of course fine with the deletes here @yselkowitz but now looking at this I see that  we are not installing any of the templates from https://github.com/sclorg/nodejs-ex since we have no tags for their remaining nodejs.json, where there is not DB

Now, yes, I do see https://github.com/nodeshift-starters/nodejs-rest-http-crud/tree/main/.openshift and in fact I see you've been contributing to that one

Are you comfortable "speaking for SCL" and saying that https://github.com/nodeshift-starters/nodejs-rest-http-crud/tree/main/.openshift and its use of postrgesql is a suitable replacement for showing nodejs work with a DB, and bypassing sclorg/nodejs-ex?

Full disclosure: I have not had the time to compare those repos in detail for similarities and differences

Or do we need to confer with them

yselkowitz: This was coordinated with our Node.js and SCL teams, and it was Node.js that chose to replace nodejs-ex with nodejs-rest-http-crud.  If someone goes through the work of migrating nodejs-ex to a different database, then that too could be considered as an addition, but that hasn't happened.

gabemontero: OK so we are in sync with them ... thanks, just needed to do a sanity check.  I was probably made aware of all that some point in the past but that data got swapped out of my memory buffer

PR #283, opened 2021-07-28
[KIECLOUD-568] - Update RHPAM and RHDM templates and imagestreams on …

msmagnanijr:
…OpenShift Service Catalog for 7.11.1 release

Signed-off-by: Mauricio Magnani <msmagnanijr@gmail.com>

### DIFF
--- a/official/rhdm/imagestreams/rhdm-kieserver-rhel8.json
+++ b/official/rhdm/imagestreams/rhdm-kieserver-rhel8.json
@@ -15,17 +15,17 @@
 		},
 		"tags": [
 			{
-				"name": "7.11.0",
+				"name": "7.11.1",
 				"annotations": {
 					"description": "Red Hat Decision Manager 7.11 - KIE Server image.",
 					"iconClass": "icon-jboss",
 					"supports": "rhdm:7.11",
 					"tags": "rhdm",
-					"version": "7.11.0"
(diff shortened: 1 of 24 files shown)


msmagnanijr:
@bparees can you please take a look? thanks!

msmagnanijr:
> @bparees can you please take a look? thanks!

Maybe @adambkaplan you can take a look as well :) thanks!

official/rhdm/imagestreams/rhdm-kieserver-rhel8.json
```diff
@@ -15,17 +15,17 @@
 		},
 		"tags": [
 			{
-				"name": "7.11.0",
+				"name": "7.11.1",
```
bparees, line 18: we typically try not to remove tags, but rather just add new ones (and potentially have a `latest` tag that aliases the latest version, for people who want to stay on the bleeding edge) so that we do not break people who have refs to the old tag names.

spolti: Hi @bparees, we usually don't keep prior versions when a new is released. Our templates are deprecated already, we just keep them around if some customer wants to try it, however if they need older version they will grab it from our repositories directly. We've been doing it for a while now, since 7.5.0 IIRC.

PR #289, opened 2021-09-24
[KEYCLOAK-18234] Update openshift/library with the RH-SSO 7.5.0.GA te…

sguilhen:
…mplates and imagestreams


dperaza4dustbit (approved):
/lgtm

PR #301, opened 2022-03-22
Removing JWS openj9 references, as this image has been deprecated

bekionel:
The JWS OpenJ9 image has been deprecated, so removing the relevant entries from official.yaml .


yselkowitz (changes requested):
@bekionel thanks for the PR.  This would result in nothing being available for P&Z, but aren't the 5.6 openjdk11 images now available thereon?  If so, then we also need to add the following tags to 5.6 openjdk11 imagestream/templates:
```
          - arch_x86_64
          - arch_s390x
          - arch_ppc64le
```

yselkowitz:
Thanks, a couple remaining issues:

- `webserver5el8_version` should be updated for 5.6.1
- even with that, there's a problem that the imagestream still contains a 5.6.0 tag which is x86_64 only, which openshift won't like because the pull will fail.  Could the 5.6.0 tag be removed?  If not, I suppose we would need P&Z-specific imagestreams with only 5.6.1+, which would be an additional maintenance burden on you. :-1:

bekionel:
@yselkowitz thank you for your comments above. I've cut a new tag in the JWS openshift image repo where the 5.6.0 tag is removed and I've updated the webserver5el8_version variable to reflect that. Please let me know if all is in order now, cheers!

PR #303, opened 2022-03-29
Bump dependencies to address protobuf security issues

bparees:


bparees:
@dperaza4dustbit @coreydaley ptal, addresses https://github.com/openshift/library/security/dependabot/1

PR #315, opened 2022-05-23
RHMWRT-96 - Drop old EAP imagestreams and templates

luck3y:


yselkowitz:
@luck3y thanks for your PR.  #313 cleaned up the old EAP versions along with other runtimes, so I think we're set now.
/close

luck3y:
Thanks Yaakov!

PR #316, opened 2022-05-24
Enable EAP for ppc64le and s390x

yselkowitz:
These architectures are now supported in the openjdk11 based images.

/cc @luck3y


luck3y:
+1 thanks @yselkowitz, I didn't know that was specified in here.

dperaza4dustbit (approved):
/lgtm

PR #317, opened 2022-05-24
Remove unused JWS 3.1 templates

yselkowitz:
JWS 3.1 remains supported but these particular templates are not, and have not been shipped in OCP since 4.10.

/cc @bekionel


dperaza4dustbit (approved):
/lgtm

PR #321, opened 2022-06-08
Cleanup fuse samples

yselkowitz:
Fuse 7.0 is no longer supported.  Fuse 6.3 remains supported but was never supported on OCP 4.

/assign @dperaza4dustbit 
/cc @fbm3307


dperaza4dustbit (approved):
/lgtm

PR #331, opened 2022-07-20
[RHSSO-2058] Add imagestream & templates definition for RH-SSO "v7.6.0.GA" tag (a replacement for former #329 PR)

iankko:
[RHSSO-2058] Add imagestream & templates definition for "v7.6.0.GA"
    tag / release of Red Hat Single Sign-On 7.6 for OpenJDK on OpenShift
    multiarch container image. Relevant advisory ID is: RHEA-2022:5453
    
    Signed-off-by: Jan Lieskovsky <jlieskov@redhat.com>

**Note:**

This is a replacement of the former #329 one to start using latest `13-el8` tag of PostgreSQL
container image instead of the deprecated `10` tag.

@yselkowitz @fbm3307 @dperaza4dustbit PTAL once got a chance

Thx!
Jan

### DIFF
--- a/official.yaml
+++ b/official.yaml
@@ -3,6 +3,7 @@ variables:
   openjdk_version: release
   amq_version: ose-v1.4.18
   rhsso75_openjdk_version: v7.5.2.GA
+  rhsso76_openjdk_version: v7.6.0.GA
   webserver_version: ose-v1.4.17
   webserver3_version: jws31-v1.4
   webserver5el8_version: jws56el8-v5.6.2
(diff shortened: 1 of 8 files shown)


fbm3307:
@iankko I would like to suggest that instead of creating templates in openshift/library, you can create helm charts for the same (See here https://github.com/openshift-helm-charts/charts )

iankko:
> @iankko I would like to suggest that instead of creating templates in openshift/library, you can create helm charts for the same (See here https://github.com/openshift-helm-charts/charts ) It is easier to create helm charts and also the helm charts would be available across ocp version 4.8 to latest 4.11. In this you will not have to worry about backporting or waiting for the release, also no worries on checking if compatible with the new release and all those dependencies.

@fbm3307 Thank you for the suggestion, Feny! FWIW regarding helm charts I can only second to your suggestion, but RH-SSO templates have been there for pretty long time (for the time being starting of RH-SSO 7.0.X for the very least AFAICT). So the need to update the templates as part of each RH-SSO release is some way of backward compatibility for the customers (for those already accustomed to expected new templates version with a new release, providing just helm charts, they might need to change their workflow too etc.)

We can consider providing Helm charts too, but not completely sure if it would be just replacement (replace former templates with new Helm charts). Definitely something for further discussion / consideration. But we will consider this for future releases.

iankko:
@fbm3307 @gabemontero By any chance could this be reviewed / approved soon? (it has been quite some time from the original `7.6.0.GA` release. ATM I am not aware of any customer requests regarding 7.6.0 templates, but in any case the review process seemed to be a bit more prompt in the past AFAICT. Maybe it's just due to a PTO season).

Thanks for your time & look in advance in any case!

gabemontero:
/assign @fbm3307 
/assign @dperaza4dustbit 

fyi @iankko I've got a new assignment, and won't be processing these PRs

iankko:
@fbm3307 

Thank you very much!

PR #332, opened 2022-07-20
Add build template for OpenJDK to IBM Semeru transition

dale-fu:
Initial PR to add generic template to convert OpenJDK 11 to IBM Semeru 11.

### DIFF
--- a/official.yaml
+++ b/official.yaml
@@ -1045,3 +1045,11 @@ data:
         tags:
           - ocp
           - arch_aarch64
+  semeru:
+    templates:
+      - location: https://raw.githubusercontent.com/ibm-redhat-mw-PE/openjdk-semeru-transition/main/openjdk-to-semeru-transition.json
+        docs: https://github.com/ibm-redhat-mw-PE/openjdk-semeru-transition/blob/main/README.md
+        tags:
+          - ocp
+          - arch_s390x
+          - arch_ppc64le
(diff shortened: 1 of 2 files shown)


fbm3307:
@dale-fu  can you shed some more light onto this for us to understand better , as to why this is required

levivic:
hi @fbm3307, the template is needed for potential customers to switch their JVM choice of certain RH openshift products which are based on Java, such as EAP, RHSSO .... Details can be found in https://access.redhat.com/articles/6744521. This template should be part of the last section of the article, `Transitioning back to OpenJ9`.  After some discussion with affected products teams, we agree to use openshift/library to host the template.

fbm3307:
@dale-fu @levivic 
I would like to suggest that instead of creating templates in openshift/library, you can create helm charts for the same (See here https://github.com/openshift-helm-charts/charts )

levivic:
@fbm3307 thanks for the info. As mentioned in my email reply, we have already started working with the doc team to make it clear to the clients and we will direct them to the openshift/library github repo directly once it's merged. Also due to the urgency of getting the solution delivered to customers (end of the month) and other prioritized tasks the team has, we will look at the chart based solution later, after the original template solution gets published.

fbm3307:
Hi @levivic , Since this points to location of a personal repo, as per OCP guidelines, this PR cannot be merged.

levivic:
@fbm3307 what the guidelines are? what kind of repo would be qualified?

levivic:
@fbm3307 could you please merge the PR as @dale-fu already moved the original template to a public org?

fbm3307:
@levivic done

PR #333, opened 2022-08-01
[KIECLOUD-586] - Update RHPAM and RHDM templates and imagestreams on …

spolti:
…OpenShift Service Catalog for 7.13.0 release


fbm3307:
Hi @spolti I would like to suggest that instead of creating templates in openshift/library, you can create helm charts for the same (See here https://github.com/openshift-helm-charts/charts )

spolti:
Hi @fbm3307 we do not maintain these templates anymore, we still have it because we have a few customers that are still using it on ocp 3.11 and have no plans to create helm charts for it, but, this repo is for ocp 4.x, right? If so I'll close it. :)

fbm3307:
Hi @spolti yes, and can these be removed if they are not supported in 4.x ? if yes can you raise a PR for the same?

spolti:
@fbm3307 thanks for confirming, about the previous versions, do we really need to remove it or do you mean just remove it from the official.yaml file?

fbm3307:
@spolti we need to remove all of it from jsons.. and also remove it from library.( whichever are no longer supported)

PR #350, opened 2023-08-31
community.yaml: Update sources to C9S/UBI9

LorbusChris:
Removes stale and deprecated sources


vrutkovs (approved):

fbm3307:
Taking a look at this PR

LorbusChris:
Note that the only manual change here was to `community.yaml` - the rest of the changes is generated from that.

fbm3307:
@LorbusChris yes , but i do need to make sure it doesn't interfere with Cluster Samples operator, as CSO is currently being downsized and hence the delay.

LorbusChris:
There are further PRs which will affect what gets pulled into OpenShift:
- https://github.com/sclorg/httpd-container/pull/191
- https://github.com/sclorg/mariadb-container/pull/216
- https://github.com/sclorg/mysql-container/pull/349

fbm3307:
@LorbusChris will be approving  this PR, but just FYI, that we are downsizing cluster samples operator, The phase-1 of it is already in progress and by year end we will pull out the support for Non-s2i samples and then there are plans for converting S2i ones to helmcharts. so CSO will stop managing the imagestreams/templates. so if you are updating this for long term basis you might want to consider other option if these are meant for openshift.

LorbusChris:
@fbm3307 thanks for the context. I'm not really attached to the samples operator, so all that's fine with me.

This change is mainly to get rid of references to docker.io in the samples we ship today, as those make us hit rate limits in CI.

fbm3307:
@LorbusChris okay makes sense

